PR #2, opened 2023-07-21
EF-18: Remove deleted entities from database on save.

rstam:


src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -63,20 +64,8 @@ public class MongoDatabaseWrapper : Database
...
-            // TODO: Exception handling
-            if (Save(entry))
-            {
-                docsAffected++;
-            }
-        }
-
-        return docsAffected;
+        var updates = ConvertUpdateEntriesToMongoUpdates(entries);
+        return (int)SaveMongoUpdates(updates);
```
rstam, line 68: At a high level saving changes is merely a matter of converting the EF Core update entries to MongoDB updates and running a bulk update.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -87,20 +76,8 @@ public override int SaveChanges(IList<IUpdateEntry> entries)
...
-            // TODO: Exception handling
-            if (await SaveAsync(entry, cancellationToken))
-            {
-                docsAffected++;
-            }
-        }
-
-        return docsAffected;
+        var updates = ConvertUpdateEntriesToMongoUpdates(entries);
+        return (int)await SaveMongoUpdatesAsync(updates, cancellationToken).ConfigureAwait(false);
```
rstam, line 80: All uses of `await` should have `ConfigureAwait(false)`.

Consider adding the analyzer we use in the driver to enforce this.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -217,4 +148,148 @@ private static string GetId(IUpdateEntry entry)
             _ => throw new NotSupportedException($"Unknown type for _id property on entity type ${entry.EntityType.DisplayName()}")
         };
     }
+
+    private IEnumerable<MongoUpdate> ConvertUpdateEntriesToMongoUpdates(IEnumerable<IUpdateEntry> entries)
+    {
+        return
+            GetAllChangedRootEntries(entries)
```
rstam, line 155: I'm not 100% sure of `GetAllChangedRootEntries` but for the purposes of this PR I am assuming it is correct.

Example question: if a nested document has a state of `Modified` does that automatically mean the root document also has a state of `Modified`? If not the root document won't get saved.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -217,4 +148,148 @@ private static string GetId(IUpdateEntry entry)
...
+            .Select(ConvertUpdateEntryToMongoUpdate)
+            .Where(x => x != null)
+            .Cast<MongoUpdate>();
```
rstam, line 158: The `Cast` is to shut up a compiler warning about nullability. Apparently the compiler doesn't realize that `Where(x => x != null)` guarantees there won't be any null values.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -217,4 +148,148 @@ private static string GetId(IUpdateEntry entry)
...
+    {
+        throw new NotImplementedException(); // EF-15
+    }
+
+    private long SaveMongoUpdates(IEnumerable<MongoUpdate> updates)
+    {
+        var database = _mongoClient.Database;
+        var client = database.Client;
+        using var session = client.StartSession();
+        return session.WithTransaction((session, cancellationToken) => SaveMongoUpdates(session, database, updates));
```
rstam, line 199: We don't have to use a transaction if we decide not to.

But this shows that it is trivial to use a transaction for the whole batch update if we want to.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -217,4 +148,148 @@ private static string GetId(IUpdateEntry entry)
...
+    }
+
+    private long SaveMongoUpdates(IClientSessionHandle session, IMongoDatabase database, IEnumerable<MongoUpdate> updates)
+    {
+        var documentsAffected = 0L;
+        foreach (var batch in BatchUpdatesByCollection(updates))
```
rstam, line 205: Since the current MongoDB bulk write API only supports bulk writes to a single collection we have to break up the updates by collection.

I'm assuming order matters and therefore not reordering updates (so different updates to the same collection will end up in different batches if interrupted by updates to other collections).

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -217,4 +148,148 @@ private static string GetId(IUpdateEntry entry)
...
+            }
+        }
+
+        if (batch != null)
+        {
+            yield return batch;
+        }
+    }
+
+    private class MongoUpdate
```
rstam, line 263: This class represents a single update to a particular collection.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -217,4 +148,148 @@ private static string GetId(IUpdateEntry entry)
...
+        {
+            CollectionName = collectionName;
+            Model = model;
+        }
+
+        public string CollectionName { get; }
+        public WriteModel<BsonDocument> Model { get; }
+    }
+
+    private class MongoUpdateBatch
```
rstam, line 277: This class represents a batch of updates to a particular collection.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -217,4 +148,148 @@ private static string GetId(IUpdateEntry entry)
...
+            GetAllChangedRootEntries(entries)
+            .Select(ConvertUpdateEntryToMongoUpdate)
+            .Where(x => x != null)
+            .Cast<MongoUpdate>();
+    }
+
+    private MongoUpdate? ConvertUpdateEntryToMongoUpdate(IUpdateEntry entry)
+    {
+        var collectionName = entry.EntityType.GetCollectionName();
+        var id = GetId(entry);
```
rstam, line 164: Note: `GetId` assumes the `_id` is going to be a string.

Leaving that assumption alone for now.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -217,4 +148,148 @@ private static string GetId(IUpdateEntry entry)
...
+    }
+
+    private MongoUpdate ConvertAddedEntryToMongoUpdate(string collectionName, string id, IUpdateEntry entry)
+    {
+        throw new NotImplementedException(); // EF-17
+    }
+
+    private MongoUpdate ConvertDeletedEntryToMongoUpdate(string collectionName, string id, IUpdateEntry entry)
+    {
+        var filter = new BsonDocument("_id", id);
```
rstam, line 184: If we decide to support types beside `string` for the `_id` then we will need to find the correct serializer to serialize the `id` value.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: Yes, that's exactly what it does - pushes the change indicator on embedded owned/entities up to the root/owned entity.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: Let's keep the transactions.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: Yeah I think we should do that next if you don't want to include it in this piece of work.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
rstam: Is there a bug on this line then?

https://github.com/mongodb/mongo-efcore-provider/blob/main/src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs#L121

Seems like it should use `==` instead of `!=`.

rstam: Question about this line also:

https://github.com/mongodb/mongo-efcore-provider/blob/main/src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs#L127

Won't the `root` end up in the `HashSet` twice? Once `Unchanged` and again `Modified`? Similarly, couldn't the `root` end up in the `HashSet` multiple times (all as `Modified`) if the same root entity has more than one modified owned type?

rstam:
@damieng can you help me figure out how to write tests?

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
ahmednfwela: tip: you can shortcut both lines into just `OfType`
```diff
    .Select(ConvertUpdateEntryToMongoUpdate)
+    .OfType<MongoUpdate>();
-    .Where(x => x != null)
-    .Cast<MongoUpdate>();
```

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
ahmednfwela: will this also work in cases where it's just a single node? (not a replica set)

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
ahmednfwela: you can also use [`Fody.ConfigureAwait`](https://github.com/Fody/ConfigureAwait) to automate this

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: A hashset does not allow duplicates so that is fine.

We only want to change the root to modified if it was unmodified. We do not want to change it from added or deleted to modified (which the Cosmos provider does but is a bug afaik)

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: Ordering should not matter in our provider as we don't have FK constraints, cascading deletes etc. that SQL has to worry about.

damieng:
Are you talking about unit tests or full functional tests? We have a few of the former already checked in so there's a pattern there. We don't however have any end-to-end testing yet. We could either do functional testing with a spun-up test mongo instance or mock out some of the lower layers of the provider to intercept and capture/replay. What's your preference?

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
rstam: I did notice that it was a HashSet...

So are you assuming that the `IUpdateEntry` implementation of `Equals` uses reference equality?

Are you assuming that changing the state of an `IUpdateEntry` (by changing the `EntityState` property) doesn't affect the results of `Equals` and `GetHashCode`?

Using HashSet with mutable types is perilous.

rstam:
Mocking lower layers is a pain to do and isn't a very useful test in my opinion.

The driver integration tests assume an instance of MongoDB is running at `localhost:27017`. I always have such an instance of MongoDB running on my machine at all times. I suspect everyone on the team does.

Testing with a real instance of MongoDB ensures that our interactions with MongoDB are behaving as they should.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: That is true but it is fine here - the Cosmos provider and EF internals use HashSet's of IUpdateEntity and the only implementation has neither Equals nor GetHashCode defined so it is all reference based on the mutable types.

damieng:
Sounds fine to me - do you want to add a FunctionalTests project and put some plumbing in to exercise it? If you want to work together I can pair up on Monday.

damieng (approved):
LGTM

PR #3, opened 2023-07-25
EF-20: Initial minimal version of EntitySerializer.

rstam:


damieng (approved):
Looks like a good start. We can enhance the sections as we enable more fine-grained model configuration.

PR #6, opened 2023-08-04
EF-17: Initial version of handling Insert.

rstam:


src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
```diff
@@ -22,6 +22,16 @@
 
 namespace MongoDB.EntityFrameworkCore.Serializers
 {
+    internal static class EntitySerializer
+    {
+        public static IBsonSerializer Create(IReadOnlyEntityType entityType)
```
rstam, line 27: This helper method makes it easier to instantiate instances of `EntitySerializer<TValue>` when `TValue` is not known until runtime.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -187,7 +191,41 @@ private IEnumerable<MongoUpdate> ConvertUpdateEntriesToMongoUpdates(IList<IUpdat
 
     private static MongoUpdate ConvertAddedEntryToMongoUpdate(string collectionName, string id, IUpdateEntry entry)
     {
-        throw new NotImplementedException(); // EF-17
+        var entityType = entry.EntityType;
```
rstam, line 194: I'm still confused about which interfaces we are supposed to use.

I thought we were supposed to use `IReadOnlyEntityType`, but `entry.EntityType` is of type `IEntityType`.

I guess we will continue to use `IReadOnlyEntityType` (instead of `IEntityType`) whenever possible?

Everything still compiles because `IEntitType` implements `IReadOnlyEntityType`.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -187,7 +191,41 @@ private IEnumerable<MongoUpdate> ConvertUpdateEntriesToMongoUpdates(IList<IUpdat
...
+        var entitySerializer = (IBsonDocumentSerializer)EntitySerializer.Create(entityType);
+
+        var document = new BsonDocument();
+        using (var writer = new BsonDocumentWriter(document))
+        {
+            writer.WriteStartDocument();
+
+            foreach (var property in entityType.GetProperties())
```
rstam, line 202: I'm surprised we can't get the whole instance to serialize, but it is what it is I guess.

Looks like we are forced to serialize the properties one at a time instead of serializing the whole document.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -187,7 +191,41 @@ private IEnumerable<MongoUpdate> ConvertUpdateEntriesToMongoUpdates(IList<IUpdat
 
     private static MongoUpdate ConvertAddedEntryToMongoUpdate(string collectionName, string id, IUpdateEntry entry)
     {
-        throw new NotImplementedException(); // EF-17
+        var entityType = entry.EntityType;
+        var entitySerializer = (IBsonDocumentSerializer)EntitySerializer.Create(entityType);
```
rstam, line 195: This is the serializer that will determine what the documents in the database look like.

This is the same serializer we pass to the LINQ provider (it *has* to be the same one, or LINQ queries won't get translated correctly).

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -187,7 +191,41 @@ private IEnumerable<MongoUpdate> ConvertUpdateEntriesToMongoUpdates(IList<IUpdat
...
+            foreach (var property in entityType.GetProperties())
+            {
+                var propertyValue = entry.GetCurrentValue(property);
+                var propertySerializationInfo = GetPropertySerializationInfo(entitySerializer, property);
+                var elementName = propertySerializationInfo.ElementName;
+                var propertySerializer = propertySerializationInfo.Serializer;                   
+                var context = BsonSerializationContext.CreateRoot(writer);
+
+                writer.WriteName(elementName);
+                propertySerializer.Serialize(context, propertyValue);
```
rstam, line 211: You talked about converters... but I don't know what those are, and in any case in the MongoDB C# driver world it is the serializers that determine how things are stored in the database.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: IReadOnlyEntityType is a version of IEntityType without the methods/properties to modify it. Ideally after the DbContext is created everything should be IReadOnlyEntityType but it sounds like they have a few places where this isn't the case. I'd not worry about it but I can mention it on the next call with the EF team.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: The whole document serializer would surely go over the properties as well though but using reflection?

There are some properties that we wouldn't want to serialize on the Clr such as those marked [Ignore] or configured to not be mapped in EF - their GetProperties ensures we only get the properties we have been told to persist.

damieng (approved):

PR #7, opened 2023-08-04
EF-15: Initial version of updating modified entities.

rstam:


src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -199,7 +203,60 @@ private static MongoUpdate ConvertDeletedEntryToMongoUpdate(string collectionNam
...
-        throw new NotImplementedException(); // EF-15
+        var entityType = entry.EntityType;
+        var entitySerializer = (IBsonDocumentSerializer)EntitySerializer.Create(entityType);
+
+        var idProperty = entityType.GetIdProperty() ?? throw new InvalidOperationException($"Type {entityType.ClrType.Name} has no Id property.");
+        var idValue = entry.GetCurrentValue(idProperty);
+        var idSerializationInfo = GetPropertySerializationInfo(entitySerializer, idProperty);
+        var idSerializer = idSerializationInfo.Serializer;
+        var serializedIdValue = SerializeValue(idSerializer, idValue);
+        var filterDefinition = Builders<BsonDocument>.Filter.Eq("_id", serializedIdValue);
```
rstam, line 214: The code above no longer assumes that the _id property is a string.

We can extract this to a helper method and use it for deleting documents also.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -199,7 +203,60 @@ private static MongoUpdate ConvertDeletedEntryToMongoUpdate(string collectionNam
...
+
+        var fieldValues = new BsonDocument();
```
rstam, line 216: Only the properties that were modified will end up being added to the fieldValues document.

damieng (approved):
LGTM

PR #8, opened 2023-08-08
Correctly serialize the id property for deletions

damieng:
I was previously assuming that we could pass a raw value to the FilterDefinition but it needs to be correctly serialized into a BsonValue.

I've re-used the code from the update method into a new CreateIdFilter helper.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
+++ b/src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
@@ -131,24 +131,6 @@ public class MongoDatabaseWrapper : Database
     }
 
     /// <summary>
-    /// Get the unique _id for a given <see cref="IUpdateEntry"/>.
-    /// </summary>
-    /// <param name="entry">The <see cref="IUpdateEntry"/> to obtain the _id for.</param>
-    /// <returns>The _id for this entry expressed as an object.</returns>
-    /// <exception cref="InvalidOperationException">If no property could be identified on the <see cref="IUpdateEntry"/> as holding the _id.</exception>
-    private static object GetId(IUpdateEntry entry)
-    {
-        var idProperty = entry.EntityType.GetIdProperty();
-        if (idProperty == null)
-        {
-            throw new InvalidOperationException(
-                $"The entity type '{entry.EntityType.DisplayName()}' does not have a property mapped to the 'id' property in the database. Add a property mapped to 'id'");
-        }
-
-        return entry.GetCurrentProviderValue(idProperty)!; // TODO: use _id serializer from EntitySerializer
-    }
-
-    /// <summary>
     /// Convert a list of <see cref="IUpdateEntry"/> from EF core into <see cref="MongoUpdate"/> we
     /// can send to MongoDB.
     /// </summary>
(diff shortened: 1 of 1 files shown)


src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -238,57 +223,67 @@ private static MongoUpdate ConvertModifiedEntryToMongoUpdate(string collectionNa
...
         return new MongoUpdate(collectionName, model);
+    }
 
-        static BsonSerializationInfo GetPropertySerializationInfo(IBsonDocumentSerializer entitySerializer, IReadOnlyProperty property)
-        {
-            if (entitySerializer.TryGetMemberSerializationInfo(property.Name, out var serializationInfo))
-            {
-                return serializationInfo;
-            }
+    private static FilterDefinition<BsonDocument> CreateIdFilter(IBsonDocumentSerializer serializer, IUpdateEntry entry)
```
rstam, line 249: I would rename `serializer` to `entitySerializer`.

There are lots of serializers involved and it helps to know which one we are talking about.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -238,57 +223,67 @@ private static MongoUpdate ConvertModifiedEntryToMongoUpdate(string collectionNa
...
+    {
+        var idProperty = entry.EntityType.GetIdProperty() ??
```
rstam, line 251: Should `GetIdProperty` throw an exception if it can't find the id property?

That way callers don't have to check.

I think if we call `GetIdProperty` it's because we need to have it. I can't think of a case where we would want to get `null` back here.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
```diff
@@ -238,57 +223,67 @@ private static MongoUpdate ConvertModifiedEntryToMongoUpdate(string collectionNa
...
+
+        var idValue = entry.GetCurrentValue(idProperty);
+        var serializedIdValue = SerializeValue(idSerializationInfo.Serializer, idValue);
 
-        static BsonValue SerializeValue(IBsonSerializer serializer, object? value)
+        return Builders<BsonDocument>.Filter.Eq("_id", serializedIdValue);
+    }
+
+    private static BsonSerializationInfo GetPropertySerializationInfo(IBsonDocumentSerializer entitySerializer,
+        IReadOnlyProperty property)
```
rstam, line 263: As a readability issue I prefer all parameters on a single line, or once you need multiple lines, one parameter per line.

At a quick glance this declaration "looks" like it only takes a single parameter called `property`.

    private static BsonSerializationInfo GetPropertySerializationInfo(
        IBsonDocumentSerializer entitySerializer,
        IReadOnlyProperty property)

rstam (changes requested):
Looks good. Just a few minor comments.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseWrapper.cs
damieng: Yeah I agree with you here, this is how Rider is formatting the file using the .editorconfig we have - I need to see if we can tweak the settings to see if we can get it to do this for us.

rstam (approved):
LGTM

PR #13, opened 2023-09-06
EF-55: Async methods to retrieve data fails to execute

sanych-sun:


damieng (approved):

damieng (changes requested):
Can we add ToListAsync and a couple of others? 

Fix was way simpler than I thought it would be, thanks for investigation and fix!

damieng (approved):

PR #14, opened 2023-09-07
Use MongoDb driver's serializers in materializer expressions.

sanych-sun:


src/MongoDB.EntityFrameworkCore/Query/Visitors/ProjectionBindingRemovingExpressionVisitorBase.cs
```diff
@@ -266,7 +266,7 @@ protected override Expression VisitMethodCall(MethodCallExpression methodCallExp
         IProperty property,
         Type type)
     {
-        var storeName = property.GetElementName();
+        var storeName = property.GetDefaultElementName();// .GetElementName();
```
sanych-sun, line 269: This is absolutely temporary code:
I would propose the approach so the materializer expression would NOT try to resolve the element name, but works with entity property names only. So it should pass property name and IEntityType down to BsonBinding... and it will resolve property or navigation property metadata from the entity type. Let's discuss it later.

src/MongoDB.EntityFrameworkCore/Query/Visitors/ProjectionBindingRemovingExpressionVisitorBase.cs
```diff
@@ -266,7 +266,7 @@ protected override Expression VisitMethodCall(MethodCallExpression methodCallExp
...
         if (storeName.Length == 0)
```
sanych-sun, line 270: If you are OK with suggestion above - we should refactor this by moving the conditions from GetDefaultElementName extension method to here.

src/MongoDB.EntityFrameworkCore/Query/Visitors/ProjectionBindingRemovingExpressionVisitorBase.cs
sanych-sun: BTW, as an additional motivation not to resolve the element name could be the "composite key" problem - when a property should be stored as "_id: { prop1: val1, prop2: val2 }" form. It might makes sense to localize this logic outside of expression building code.

damieng (approved):

PR #16, opened 2023-09-13
EF-32: Support Nullable<> data types

sanych-sun:


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ClrTypeMappingTests.cs
```diff
@@ -411,4 +418,38 @@ public void Double_negative_read()
         Assert.NotNull(actual);
         Assert.Equal(expected, actual.aDouble);
     }
+
+    [Theory]
+    [InlineData(typeof(short), (short)0)]
```
sanych-sun, line 423: Please pay your attention to this test. I suppose we can rewrite all tests in this class using this approach. The pros of it: we easily can create more test cases without necessarity to create more entity classes.

damieng: Yeah, we just have to be careful about things like boxing etc. They won't cause problems in regular code but when we're processing expression trees it's very easy to think you're covering a specific path when you're not. This one is fine.

damieng (approved):

PR #17, opened 2023-09-14
EF-43: Support Enums

sanych-sun:


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ClrTypeMappingTests.cs
```diff
@@ -434,12 +434,23 @@ public void Double_negative_read()
...
         var methodInfo = this.GetType().GetMethod(nameof(ClrTypeMappingTestImpl), BindingFlags.Instance | BindingFlags.NonPublic);
         methodInfo.MakeGenericMethod(valueType).Invoke(this, new[] { value });
     }
 
+    private enum TestEnum
+    {
+        EnumValue0 = 0,
+        EnumValue1 = 1
+    }
+
```
damieng, line 453: Can we include some test coverage for inserts of entities with Enums? 

Right now the test uses the Mongo C# Driver to insert the data for the test which is fine for exercising the query and materialization paths but insert/update path should also have coverage.

damieng (changes requested):
We should probably also add support for configuring if the enum is stored in the database as a number or a string. (Also enums don't have to be just int's so we should consider test/support for those as well)

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ClrTypeMappingTests.cs
sanych-sun: Done.

sanych-sun:
More tests added. Regarding enum representation configuration - isn't it a part of the field tuning/configuration task? https://jira.mongodb.org/projects/EF/issues/EF-44

damieng (approved):

PR #19, opened 2023-09-18
EF-33: Composite key save operation

sanych-sun:


src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
```diff
@@ -44,25 +45,63 @@ public static void WriteProperties(BsonDocument document, IUpdateEntry entry, IE
...
+        }
+
         foreach (var property in properties)
         {
-            var propertyValue = entry.GetCurrentValue(property);
-            var propertySerializationInfo = GetPropertySerializationInfo(property);
-            // TODO: Add ElementPath support here
-            var elementName = propertySerializationInfo.ElementName;
-            writer.WriteName(elementName);
+            if (property.IsPrimaryKey())
```
damieng, line 74: I personally prefer putting these into the foreach loop - e.g. `properties.Where(p => !p.IsPrimaryKey())`. Not a show stopper though.

damieng (approved):

PR #23, opened 2023-09-26
Do not write non-stored properties, split key and non-key serialization

damieng:
Splits the key and non-key parts of seialization to ensure filtering is clear and prevent writing non-stored properties.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
+++ b/src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
@@ -40,61 +40,58 @@ internal static class SerializationHelper
         return ReadElementValue<T>(document, serializationInfo);
     }
 
-    public static void WriteProperties(IBsonWriter writer, IUpdateEntry entry, Func<IProperty, bool>? propertyFilter = null)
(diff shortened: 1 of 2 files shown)


src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
```diff
@@ -40,61 +40,58 @@ public static T GetElementValue<T>(BsonDocument document, string elementName)
         return ReadElementValue<T>(document, serializationInfo);
     }
 
-    public static void WriteProperties(IBsonWriter writer, IUpdateEntry entry, Func<IProperty, bool>? propertyFilter = null)
+    internal static void WriteKeyProperties(IBsonWriter writer, IUpdateEntry entry)
```
sanych-sun, line 43: Why do we need `internal` here (and below)? the entire class is internal.

sanych-sun (approved):
LGTM + minor comment

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
damieng: Just to clarify we don't expect it to be called externally. 

The public two are the ones we bind to via reflection.

PR #25, opened 2023-09-26
EF-24: Owned entity collections

damieng:
Implementation of owned entity collections.

The main issue here was handling the synthetic non-stored keys on the collection which as well as having the parent key like non-collections also needed another value. We use a generator even though we don't persist or read these values to keep EF happy.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Extensions/MongoEntityTypeBuilderExtensions.cs
+++ b/src/MongoDB.EntityFrameworkCore/Extensions/MongoEntityTypeBuilderExtensions.cs
@@ -1,19 +1,20 @@
 ﻿/* Copyright 2023-present MongoDB Inc.
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with the License.
-* You may obtain a copy of the License at
-*
-* http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing, software
-* distributed under the License is distributed on an "AS IS" BASIS,
-* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-* See the License for the specific language governing permissions and
-* limitations under the License.
-*/
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
 using System;
+using Microsoft.EntityFrameworkCore.Diagnostics;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using MongoDB.EntityFrameworkCore.Metadata;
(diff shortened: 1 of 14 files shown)


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/OwnedEntityTests.cs
```diff
@@ -96,12 +96,14 @@ public void OwnedEntity_nested_one_level_creates()
         }
     }
 
-    [Fact(Skip = "We do not yet support updating owned types")]
-    public void OwnedEntity_nested_collection_creates()
+    [Fact]
```
sanych-sun, line 99: We probably need another test for the case when an entity has 2 navigation properties of same type. Like: 
```
class Person
{
  public Location PrimaryLocation { get; set; }
  public Location SecondaryLocation { get; set; }
}
```

tests/MongoDB.EntityFrameworkCore.UnitTests/ValueGeneration/MongoValueGeneratorSelectorTests.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.ValueGeneration;
+using MongoDB.Bson;
+using MongoDB.EntityFrameworkCore.ValueGeneration;
+
+namespace MongoDB.EntityFrameworkCore.UnitTests.ValueGeneration;
+
+public class MongoValueGeneratorSelectorTests
+{
+    private readonly IModel __model;
```
sanych-sun, line 26: Just curious what double-underscores means here?

tests/MongoDB.EntityFrameworkCore.UnitTests/ValueGeneration/OwnedEntityIndexValueGeneratorTests.cs
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+using MongoDB.EntityFrameworkCore.ValueGeneration;
+
+namespace MongoDB.EntityFrameworkCore.UnitTests.ValueGeneration;
+
+public class OwnedEntityIndexValueGeneratorTests
+{
+    [Fact]
+    public void Next_generates_non_temporary_id()
```
sanych-sun, line 23: Should we add tests to ensure no duplicates generated for concurrent execution?

src/MongoDB.EntityFrameworkCore/Extensions/MongoEntityTypeBuilderExtensions.cs
```diff
@@ -115,4 +116,77 @@ public static class MongoEntityTypeBuilderExtensions
...
+    /// <returns>The same builder instance if the configuration was applied, <see langword="null" /> otherwise.</returns>
+    public static IConventionEntityTypeBuilder? SetElementName(
+        this IConventionEntityTypeBuilder entityTypeBuilder,
+        string? name,
+        bool fromDataAnnotation = false)
+    {
+        if (!entityTypeBuilder.CanSetElementName(name, fromDataAnnotation))
+        {
+            return null;
+        }
```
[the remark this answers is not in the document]
damieng: It does seem odd but that's what the Cosmos one does https://github.com/dotnet/efcore/blob/release/8.0/src/EFCore.Cosmos/Extensions/CosmosPropertyBuilderExtensions.cs#L80 - frankly I'd have expected it to throw...

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/OwnedEntityTests.cs
damieng: Can't add this one yet as it would fail - it's a known bug EF-58.

tests/MongoDB.EntityFrameworkCore.UnitTests/ValueGeneration/MongoValueGeneratorSelectorTests.cs
damieng: Double-underscore is private static which comes from the naming rules in the .editorconfig we borrowed from the Mongo C# Driver repository. Here however their name should have been corrected when they stopped being static, will fix.

tests/MongoDB.EntityFrameworkCore.UnitTests/ValueGeneration/OwnedEntityIndexValueGeneratorTests.cs
damieng: We could but we don't currently have any of that level of test coverage and this interlock pattern is what they use throughout EF Core.

src/MongoDB.EntityFrameworkCore/Extensions/MongoEntityTypeBuilderExtensions.cs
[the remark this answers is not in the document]
AndriySvyryd: `IConvention*Builder` types are not Fluent API and returning `null` is expected in cases where an overriding configuration is present.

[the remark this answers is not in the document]
sanych-sun: It might be, but from user prospective: this is extension method to IConventionPropertyBuilder which returns IConventionPropertyBuilder... it looks like I can chain more configuration code (that's why I see this as fluent). And if I chain more code, it might result to NullReferenceException, instead of some more reasonable exception. So we might want to either to change it to be void or throw there for consistency.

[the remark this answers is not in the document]
AndriySvyryd: Whenever one of these methods returns `null` further configuration should be avoided, so the calls should be chained using `?.` instead of `.`

PR #27, opened 2023-09-27
EF-36: Query against composite key components

sanych-sun:


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/CompositeKeyQueryTests.cs
```diff
@@ -0,0 +1,85 @@
...
+    public void Should_query_by_key_components()
+    {
+        var dbContext = CreateContext();
+
+        var result = dbContext.Entitites.Where(e => e.Key1 == "two" && e.Key2 == 3).ToList();
+
+        Assert.Single(result);
+        Assert.All(result, e => Assert.Equal("two", e.Key1));
+        Assert.All(result, e => Assert.Equal(3, e.Key2));
+    }
```
damieng, line 51: Do we support looking up key elements with non-equals operations?  e.g. e.Key2 > 3 etc.

PR #29, opened 2023-09-29
Support EF's Find operator that uses Property(p, name)

damieng:
Includes tests for single and composite primary keys.


sanych-sun (approved):
LGTM

PR #30, opened 2023-10-03
EF-58 fix multiple owned entities

damieng:
Allows an entity containing two or more properties with owned entities to be materialized and stored.


sanych-sun (approved):
LGTM

PR #34, opened 2023-10-13
Support .NET 6

damieng:
It's low-effort to support .NET 6, it is what EF Core 7 supports and it is an LTS so worth supporting I think.


roji:
Note that while .NET 6 is LTS, EF Core 7 (which is what the Mongo provider is built against) is not... As a STS release, EF Core 7 will be out of support in May of 2024 (and .NET 6 in November).

So I personally wouldn't spend any time on it...

PR #39, opened 2023-11-07
EF-70: Let user configure the DateTimeKind for DateTime properties

sanych-sun:


src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
```diff
@@ -148,6 +148,22 @@ private static IBsonSerializer CreateTypeSerializer(Type type)
         return serializer;
     }
 
+    private static IBsonSerializer CreateDateTimeSerializer(IReadOnlyProperty? property)
+    {
+        var dateTimeKind = DateTimeKind.Unspecified;
+        if (property != null)
+        {
+            dateTimeKind = property.GetDateTimeKind();
+        }
```
damieng, line 157: var dateTimeKind = property?.GetDateTimeKind() ?? DateTimeKind.Unspecified; perhaps?

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
```diff
@@ -148,6 +148,22 @@ private static IBsonSerializer CreateTypeSerializer(Type type)
...
+
+        if (dateTimeKind == DateTimeKind.Unspecified)
+        {
+            return new DateTimeSerializer();
```
damieng, line 161: Something about the default for DateTimeSerializer being DateTimeKind.Unspecified while the convention defaults to DateTimeKind.Utc is feeling off. Perhaps we should make the serializer constructor require a DateTimeKind and remove this block?

src/MongoDB.EntityFrameworkCore/Metadata/Conventions/DateTimeKindConvention.cs
```diff
@@ -0,0 +1,64 @@
...
+/// <summary>
+/// A convention that set DateTimeKind for DateTime properties.
+/// </summary>
+public class DateTimeKindConvention : IPropertyAddedConvention
+{
+    /// <summary>
+    /// Creates a <see cref="DateTimeKindConvention"/>.
+    /// </summary>
+    /// <param name="dateTimeKind">DateTimeKind to use</param>
+    public DateTimeKindConvention(DateTimeKind dateTimeKind = DateTimeKind.Utc)
```
damieng, line 32: Wonder if we should make this require a DateTimeKind and not default?

damieng (changes requested):

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
sanych-sun: We still need this block, because passing `DateTimeKind.Unspecified` will works same way as Local, but parameterless constructor of `DateTimeSerializer` uses Utc as default value. I would prefer to follow the MongoDB driver's defaults.

damieng (approved):

PR #41, opened 2023-11-09
EF-67: KeyNotFoundException does not include key name in SerializationHelper.ReadElementValue

sanych-sun:


src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
```diff
@@ -31,13 +31,33 @@ internal static class SerializationHelper
...
+        {
+            return value;
+        }
+
+        if (property.IsNullable)
+        {
+            return default;
+        }
+
+        throw new KeyNotFoundException($"Document does not contain value for non nullable field '{property.Name}'.");
```
damieng, line 44: Should be "non-nullable" (hyphenated)

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
```diff
@@ -31,13 +31,33 @@ internal static class SerializationHelper
     public static T GetPropertyValue<T>(BsonDocument document, IReadOnlyProperty property)
     {
         var serializationInfo = GetPropertySerializationInfo(property);
-        return ReadElementValue<T>(document, serializationInfo);
+        if(TryReadElementValue(document, serializationInfo, out T value))
+        {
+            return value;
+        }
+
+        if (property.IsNullable)
```
damieng, line 39: Given that TryReadElementValue sets value to default in the false case we could inline this with the other condition,

```
if (TryReadElementValue(document, serializationInfo, out T value) || property.IsNullable) {
  return value;
}

throw...
```

damieng (changes requested):
Looks great, just a couple of nitpicks.

damieng (approved):

PR #43, opened 2023-11-09
EF-74: Camel case convention for all element names

damieng:
While the camel case convention provider is not switched on by default it is available and in preview-1 it only applied to properties and not relationships that are owned entities.

As both of these are now supported the name is updated to reflect it applies to all element names and not just those from properties.


sanych-sun (approved):
LGTM

PR #45, opened 2023-11-23
Clean up code

damieng:
While working on select rewriting I found a few areas to improve first:

- Fix `ProjectionBindingExpression` to use original type not nullable type (fixes array projections)
- Remove unnecessary processing of `Object.Equals` inside `Visit`
- Push `MemberAccess` expressions from `Visit` down into `VisitMemberAccess` for clarity
- Removed unused `_collectionShaperMapping`
- Property name evaluation moved to outside of predicate to prevent multi-evaluation


sanych-sun (approved):
LGTM

PR #47, opened 2023-12-18
EF-63: Allow filtering against sub-collections

damieng:
Fixes EF-63 which prevented queries against elements on items in non-primitive sub-collections of an entity.

This change requires https://github.com/mongodb/mongo-csharp-driver/pull/1226 to work.

Until then we need to hold off on merging this. Will also need to revert the changes to the .csproj to go to the new NuGet package rather than a local project ref that is currently checked in.


src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
```diff
@@ -66,7 +66,9 @@ public bool TryGetMemberSerializationInfo(string memberName, out BsonSerializati
             if (navigation != null)
             {
                 var entityType = navigation.TargetEntityType;
-                var serializer = _entitySerializerCache.GetOrCreateSerializer(entityType);
+                var serializer = navigation.IsCollection
+                    ? new CollectionSerializationProvider().GetSerializer(navigation.ClrType)
```
JamesKovacs, line 70: Why do we use a serializer cache for entities, but not for collection serializers?

JamesKovacs (approved):
LGTM. One minor question that doesn't need to hold up this PR. It can either be a follow-on optimization or ignored as appropriate.

src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
damieng: I'll look into unifying the cache for all types.

PR #51, opened 2024-01-22
Nullable owned entity support

damieng:
Fixes #90 and introduces two concepts:

1. Owned Entity navigations can be required or optional
2. Relies on EF metadata to tell us which (`IsRequired` method, conventions etc).
3. If sub-document is missing in the BsonDocument then either we throw a helpful error (when required) or defaults to null (when optional)

Also cleaned up some code and added a bunch of XML documentation in these areas. Some reformatting happening now have switched the editor to actually apply the checked-in .editorconfig for styling.


src/MongoDB.EntityFrameworkCore/Query/Expressions/EntityProjectionExpression.cs
```diff
@@ -58,53 +74,72 @@ public Expression BindNavigation(INavigation navigation)
...
     public Expression? BindMember(
         string name,
         Type entityType,
         out IPropertyBase? propertyBase)
         => BindMember(MemberIdentity.Create(name), entityType, out propertyBase);
 
+    /// <summary>
+    /// Bind a member for this entity projection by <see cref="MemberInfo"/>.
+    /// </summary>
+    /// <param name="memberInfo"></param>
```
BorisDog, line 107: missing comment for `memberInfo`?

BorisDog (approved):
LGTM + minor comment.

PR #52, opened 2024-01-30
Support for auto-generated ObjectId

damieng:
Fixes #92

This PR includes some clean up and refactoring which has been separated out into individual commits for easy review.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Metadata/Conventions/MongoValueGenerationConvention.cs
+++ b/src/MongoDB.EntityFrameworkCore/Metadata/Conventions/MongoValueGenerationConvention.cs
@@ -1,0 +1,86 @@
+﻿/* Copyright 2024-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using Microsoft.EntityFrameworkCore.Metadata.Conventions;
+using Microsoft.EntityFrameworkCore.Metadata.Conventions.Infrastructure;
+using MongoDB.Bson;
+
+namespace MongoDB.EntityFrameworkCore.Metadata.Conventions;
+
+/// <summary>
+/// A convention that configures store value generation as <see cref="ValueGenerated.OnAdd" /> on properties that are
+/// part of the primary key and not part of any foreign keys or were configured to have a database default value.
+/// </summary>
+internal class MongoValueGenerationConvention : ValueGenerationConvention, IEntityTypeAnnotationChangedConvention
+{
+    /// <summary>
+    /// Creates a <see cref="MongoValueGenerationConvention" />.
+    /// </summary>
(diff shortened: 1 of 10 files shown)


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/AddEntityTests.cs
```diff
@@ -70,26 +70,57 @@ class MongoSpecificTypeEntity
...
+    public void Add_simple_entity_with_unset_ObjectId()
+    {
+        var collection = _tempDatabase.CreateTemporaryCollection<SimpleEntity>();
+        var dbContext = SingleEntityDbContext.Create(collection);
+
+        var expected = new SimpleEntity {name = "Not Set"};
+        dbContext.Entitites.Add(expected);
+        dbContext.SaveChanges();
+
+        Assert.Same(expected, dbContext.Entitites.First());
```
adelinowona, line 100: Probably assert that the objectId is not all zeros here.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/AddEntityTests.cs
```diff
@@ -70,26 +70,57 @@ class MongoSpecificTypeEntity
...
+    public void Add_simple_entity_with_empty_ObjectId()
+    {
+        var collection = _tempDatabase.CreateTemporaryCollection<SimpleEntity>();
+        var dbContext = SingleEntityDbContext.Create(collection);
+
+        var expected = new SimpleEntity {_id = ObjectId.Empty, name = "Empty"};
+        dbContext.Entitites.Add(expected);
+        dbContext.SaveChanges();
+
+        Assert.Same(expected, dbContext.Entitites.First());
```
adelinowona, line 118: Same as previous comment, add an assert for objectId is not all zeros.

adelinowona (approved):
LGTM with minor changes.

src/MongoDB.EntityFrameworkCore/Metadata/Conventions/MongoValueGenerationConvention.cs
```diff
@@ -0,0 +1,86 @@
+﻿/* Copyright 2024-present MongoDB Inc.
```
sanych-sun, line 1: 2024 or 2023? in Driver we use 2010 everywhere even for new files.

src/MongoDB.EntityFrameworkCore/Metadata/Conventions/MongoValueGenerationConvention.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// <param name="property">The property determine a generation strategy for.</param>
+    /// <returns>
+    /// The store <see cref="ValueGenerated"/> strategy to set for the <paramref name="property"/>
+    /// or <see langref="null"/> if no strategy should be used.
+    /// </returns>
+    protected override ValueGenerated? GetValueGenerated(IConventionProperty property)
+    {
+        if (property.DeclaringType is IConventionEntityType && property.ClrType == typeof(ObjectId))
+        {
+            var primaryKey = property.FindContainingPrimaryKey();
```
sanych-sun, line 77: It looks like we found the property as a part of PK in the line #60 and now we "searching" back for PK? Should we add extra parameter and pass PK instead?

damieng: Would have liked to have done that but unfortunately the method signatures are set by EF and not us :(

src/MongoDB.EntityFrameworkCore/Metadata/Conventions/MongoValueGenerationConvention.cs
damieng: These files were created this year so put this years date.

I don't think we can put 2010 as you can't backdate copyright and this project was only started in 2023.

sanych-sun: In Driver we use 2010 everywhere (the year when driver project was started). So should we use 2023 for EF in the same way?

PR #49, opened 2024-01-12
NOJIRA: chore: Update to dotnet 8 + ef core 8

DylanSnel:
No Jira ticket available. 

I made sure the provider runs with dotnet 8 using the latest packages. For some reason Assert.Equivalent wouldnt work anymore. I ran tests to see if my change has the correct result and it did. This felt like something out of my control, so i decided not to change that.

This is my first PR on GitHub, so if i did something wrong with this please let me know so i can learn.
I also didn't know how to update the version number of the package.

PR #55, opened 2024-01-31
Update to Entity Framework 8.0

damieng:
Rebased from #49 to make some additional tweaks.

Thanks to @dylansnel for this work in adjusting the provider to the changes in EF8 that is included in this PR


Temoxa:
Hi!

How long wait released for EfCore 8.0?

damieng:
Hi.

There won't be another release for a few more months yet - we have a number of features we need to get in first besides the EF8 support such as discriminators, projections, transactions etc.

src/MongoDB.EntityFrameworkCore/Storage/MongoTypeMappingSource.cs
```diff
@@ -76,27 +83,46 @@ public MongoTypeMappingSource(TypeMappingSourceDependencies dependencies)
...
 
         return null;
     }
+
+    public static ValueComparer? CreateComparer(CoreTypeMapping elementMapping, Type collectionType)
+    {
+        // Only single dimension arrays are supported
+        if (collectionType.IsArray && collectionType.GetArrayRank() != 1)
+        {
+            return null;
```
sanych-sun, line 115: Should we throw here?

src/MongoDB.EntityFrameworkCore/Storage/MongoTypeMappingSource.cs
```diff
@@ -76,27 +83,46 @@ public MongoTypeMappingSource(TypeMappingSourceDependencies dependencies)
...
+        }
+
+        var elementType = collectionType.TryGetItemType(typeof(IEnumerable<>))!;
+
+        return (ValueComparer?)Activator.CreateInstance(
+            elementType.IsNullableValueType()
+                ? typeof(NullableValueTypeListComparer<>).MakeGenericType(Nullable.GetUnderlyingType(elementType) ?? elementType)
```
sanych-sun, line 122: If it takes more then one ternary operator to implement a logic, I prefer to have an good-old `if` instead for better readability.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ClrTypeMappingTests.cs
```diff
@@ -447,11 +448,9 @@ public void Double_negative_read()
     [InlineData(typeof(TestByteEnum?), TestByteEnum.EnumValue1)]
     [InlineData(typeof(int[]), null)]
     [InlineData(typeof(int[]), new[] {-5, 0, 128, 10})]
-    // TODO: investigate and fix IEnumerable property support
-    // [InlineData(typeof(IEnumerable<int>), new[] { -5, 0, 128, 10 })]
     [InlineData(typeof(IList<int>), null)]
     [InlineData(typeof(IList<int>), new[] {-5, 0, 128, 10})]
-    [InlineData(typeof(ICollection<int>), new[] {-5, 0, 128, 10})]
+    [InlineData(typeof(Collection<int>), null)]
```
sanych-sun, line 453: Why do we removed test for the ICollection?

sanych-sun (changes requested):

src/MongoDB.EntityFrameworkCore/Storage/MongoTypeMappingSource.cs
damieng: With this approach it's clear that from line 120 onwards it's going to create an instance of something, and that it's dependent on whether it's a nullable value type and if not whether it is assignable or not.

With an if condition you'd have to trawl through 3-4 if statements to find out they all do the same thing with one of three different comparers. It's optimized for scanning through code quickly - the idea being that by the time you've read the first line you can decide whether this is the line you cared about or not - vs reading 15 lines of if conditions only to find they all return a valuecomparer but that wasn't what you were looking for/cared about.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ClrTypeMappingTests.cs
damieng: ICollection<T> is covered by IList<T>.

Collection<T> does not inherit from ICollection<T> and was not previously covered.

src/MongoDB.EntityFrameworkCore/Storage/MongoTypeMappingSource.cs
damieng: This block can now actually be removed as the new value comparers support multi-dimensional arrays.

If we don't support that in the C# Driver/BsonDocument then we should add detection to that as part of the model validation instead where we can throw the right error long before anything gets here.

damieng:
Had to force push to deal with merge conflicts cause by progress on main in the mean time.

PR #58, opened 2024-03-19
Handle no BSON element for nullable child docs

damieng:
Ensure that we support loading BSON documents where a child document element is entirely missing from the BSON providing the element is nullable in the entity.

Fixes EF-101


sanych-sun (approved):
LGTM

PR #59, opened 2024-03-19
Include add & delete in doc count from SaveChanges/Async.

damieng:
Previously the SaveChanges and SaveChangesAsync code only returned the number of existing documents updated and did not also include the number of documents added or deleted.

### DIFF
--- a/tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/ContextTests.cs
+++ b/tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/ContextTests.cs
@@ -1,0 +1,223 @@
+﻿/* Copyright 2023-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/ContextTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            .Select(i => new Customer("Generated " + i))
+            .ToArray();
+
+        await db.Entitites.AddRangeAsync(items);
+        await db.SaveChangesAsync();
+
+        db.Entitites.RemoveRange(items.Take(deleteCount));
+
+        Assert.Equal(deleteCount, await db.SaveChangesAsync());
+        Assert.Equal(deleteCount, await db.Entitites.CountAsync());
```
sanych-sun, line 127: I would say we should expect `items.Count() - deleteCount` here instead, because we checking number of documents left in database after the deletion.

sanych-sun (approved):
LGTM + small changes to test suggested

PR #57, opened 2024-03-14
Support more collection types including properties defined with an interface

damieng:
Fixes EF-88 and EF-98.

PR #63, opened 2024-04-10
allow models to have plain IEnumerable<T> collections

vermarine:
This patch enables models to have plain `IEnumerable<T>` properties and not just derived types (such as `List<T>`).

This is a follow-up on #57 by @damieng


damieng:
Hi there!

Could you add a test that fails without your change and passes with it?

CollectionSerializationTests is probably the place to put it.

Thanks!

damieng:
Implementing this in #69 which has the unit tests and other changes required to implement this.

PR #65, opened 2024-04-16
Add support for MQL query logging

damieng:
Fixes EF-50.

- Fair amount of logging infra as this is our first custom log
- MQL is only logged for enumerable sets not first/single (will address in another PR)
- MQL can only be logged with the parameters in place so queries only show if sensitive logging is enabled
- Logging of other operations in future PRs


JamesKovacs (approved):
LGTM

JamesKovacs:
Tests are failing because we are explicitly compiling against the 2.22.0 driver, which doesn't contain the MQL logging functionality. We need to modify the scripts to build against 2.25.0 (minimum) and latest.

PR #66, opened 2024-04-17
Clean up update pipeline prior to logging

damieng:
The update code was somewhat tangled and lead to exposure of parts we didn't really want such as IMongoDatabase on the wrapper. Instead, now:

- `MongoUpdate` - contains the collection + WriteModel for each document update & doc writing code
- `MongoBatchUpdate` - groups the MongoUpdate operations into per-collection batches
- `SerializationHelper` - loses the document writing that only MongoUpdate needed
- `IMongoClientWrapper` - stops exposing the whole of IMongoDatabase, trading it for GetCollection and SaveUpdates/Async
- `MongoClientWrapper` - implements the interface changes
- `MongoDatabaseWrapper` - Delegates out functionality to MongoUpdate and the MongoBatchUpdate

This leaves the pieces with a clear, specific focus and makes it easier for the logging of bulk operations to happen in next PR.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Storage/MongoUpdate.cs
+++ b/src/MongoDB.EntityFrameworkCore/Storage/MongoUpdate.cs
@@ -32,23 +32,29 @@ namespace MongoDB.EntityFrameworkCore.Storage;
 
 /// <summary>
 /// Represents an update to a MongoDB database by specifying the collection
-/// name and the appropriate document inside a writemodel which indicates
+/// name and the appropriate document inside a <see cref="WriteModel{BsonDocument}"/> which indicates
 /// the type of operation.
 /// </summary>
 /// <param name="collectionName">The name of the collection this update applies to.</param>
-/// <param name="model"></param>
+/// <param name="model">The <see cref="WriteModel{BsonDocument}"/> containing the update.</param>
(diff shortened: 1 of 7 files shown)


src/MongoDB.EntityFrameworkCore/Storage/MongoUpdate.cs
```diff
@@ -0,0 +1,226 @@
...
+    private static FilterDefinition<BsonDocument> CreateIdFilter(IUpdateEntry entry)
+    {
+        _ = entry.EntityType.FindPrimaryKey() ??
+            throw new InvalidOperationException($"Cannot find the primary key for the entity: {entry.EntityType.Name}");
+
+        var document = new BsonDocument();
+        using var writer = new BsonDocumentWriter(document);
+        WriteEntity(writer, entry, p => p.IsPrimaryKey());
+
+        // MongoDB require primary key named as "_id";
```
JamesKovacs, line 111: require => requires

JamesKovacs (approved):
Minor typo in comment. Discussed naming of `IMongoClientWrapper` versus `IMongoDatabaseWrapper` being backwards as the first wraps client and database whereas the second only wraps client, but that's an EF artifact and would be more confusing if we swapped the names.

LGTM (once the typo is corrected).

PR #67, opened 2024-04-19
Implement bulk write logging

damieng:
Fixes EF-110.

Provides a log message/log event for bulk write operations as we use them instead of sequential serialized individual updates.

Note that these bulk writes are batched but preserve EF-supplied ordering so you may see multiple bulk operations for the same collection name with other intermediate bulk write operations between them.

This PR also adds using type aliases for the two generic EventDefinition signatures in order to reduce noise and improve clarity.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerExtensions.cs
+++ b/src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerExtensions.cs
@@ -23,6 +23,9 @@ using MongoDB.Driver;
 
 namespace MongoDB.EntityFrameworkCore.Diagnostics;
 
+using MqlQueryEventDefinition = EventDefinition<string, CollectionNamespace, string>;
+using BulkWriteEventDefinition = EventDefinition<string, CollectionNamespace, long, long, long>;
+
 /// <summary>
 /// MongoDB-specific logging extensions.
 /// </summary>
@@ -62,6 +65,43 @@ internal static class MongoLoggerExtensions
         }
     }
 
+    public static void ExecutedBulkWrite(
+        this IDiagnosticsLogger<DbLoggerCategory.Database.Command> diagnostics,
+        TimeSpan elapsed,
+        CollectionNamespace collectionNamespace,
+        long documentsCreated,
+        long documentedDeleted,
+        long documentsModified)
+    {
+        var definition = LogExecutedBulkWrite(diagnostics);
+
+        if (diagnostics.ShouldLog(definition))
+        {
+            definition.Log(
+                diagnostics,
+                elapsed.TotalMilliseconds.ToString(),
+                collectionNamespace,
+                documentsCreated,
+                documentedDeleted,
+                documentsModified);
+        }
+
+        if (diagnostics.NeedsEventData(definition, out var diagnosticSourceEnabled, out var simpleLogEnabled))
+        {
+            var eventData = new MongoBulkWriteEventData(
+                definition,
+                ExecutedBulkWrite,
+                elapsed,
+                collectionNamespace,
+                documentsCreated,
+                documentedDeleted,
+                documentsModified,
+                diagnostics.ShouldLogSensitiveData());
+
+            diagnostics.DispatchEventData(definition, eventData, diagnosticSourceEnabled, simpleLogEnabled);
+        }
+    }
+
     private static string LoggedStagesToMql(BsonDocument[]? documents)
         => documents == null
             ? ""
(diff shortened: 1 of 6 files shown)


tests/MongoDB.EntityFrameworkCore.FunctionalTests/LoggingTests.cs
```diff
@@ -104,4 +104,67 @@ public void Query_writes_event_via_LoggerFactory_without_mql_when_no_sensitive_l
...
+        {
+            name = "Proxima Centauri d", hasRings = false, orderFromSun = -1
+        });
+
+        Assert.Equal(7, db.SaveChanges());
+
+        var log = Assert.Single(logs, l => l.Contains("MongoEventId.ExecutedBulkWrite"));
+
+        Assert.Contains("Executed Bulk Write", log);
+        Assert.Contains($"Collection='{guidesFixture.MongoDatabase.DatabaseNamespace.DatabaseName}.planets'", log);
```
JamesKovacs, line 135: I'm confused how this assertion succeeds given the spelling error in the format string.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/LoggingTests.cs
```diff
@@ -104,4 +104,67 @@ public void Query_writes_event_via_LoggerFactory_without_mql_when_no_sensitive_l
...
+
+        var logger = Assert.Single(spyLogger.Loggers, s => s.Key == "Microsoft.EntityFrameworkCore.Database.Command").Value;
+        var log = Assert.Single(logger.Records, log =>
+            log.LogLevel == LogLevel.Information &&
+            log.EventId == MongoEventId.ExecutedBulkWrite &&
+            log.Exception == null
+        ).message;
+
+        Assert.Contains("Executed Bulk Write", log);
+        Assert.Contains($"Collection='{guidesFixture.MongoDatabase.DatabaseNamespace.DatabaseName}.planets'", log);
```
JamesKovacs, line 167: Same with this assertion.

JamesKovacs (changes requested):
Minor fixes.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerExtensions.cs
```diff
@@ -96,8 +148,34 @@ private static string ExecutedMqlQuery(EventDefinitionBase definition, EventData
...
+                        LogExecuteBulkWriteString)));
+        }
+
+        return (BulkWriteEventDefinition)definition;
     }
 
     private const string LogExecutedMqlQueryString = "Executed MQL query{newLine}{collectionNamespace}.aggregate([{queryMql}])";
+
+    private const string LogExecuteBulkWriteString =
+        "Executed Bulk Write ({elapsed} ms) Collection='{colllectionNamespace}', Created={created}, Deleted={deleted}, Modified={modified}";
```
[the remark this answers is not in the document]
JamesKovacs: In our bulkWrite docs and in the C# `BulkWriteBatchResult`, we refer to them as inserted, deleted, and modified. If `Created` is more idiomatic for EF, I don't have a strong feeling about `Created` versus `Inserted`, but if there's no preference then let's stick with `Inserted`.

https://www.mongodb.com/docs/manual/core/bulk-write-operations/

src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
```diff
@@ -116,16 +122,20 @@ public long SaveUpdates(IEnumerable<MongoUpdate> updates)
     /// </summary>
     /// <param name="updates">An <see cref="IEnumerable{MongoUpdate}"/> containing the updates to apply to the database.</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
-    /// <returns>A <see cref="Task{long}"/> that when resolved gives the number of documents modified.</returns>
+    /// <returns>A <see cref="Task"/> that when resolved gives the number of documents modified.</returns>
```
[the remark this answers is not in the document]
damieng: You can only reference generics using type arguments so `Task{long}` isn't valid. The EF docs for SaveChanges doesn't even link despite returning a `Task<int>`:

```
    /// <returns>
    ///     A task that represents the asynchronous save operation. The task result contains the
    ///     number of state entries written to the database.
    /// </returns>
```

tests/MongoDB.EntityFrameworkCore.FunctionalTests/LoggingTests.cs
damieng: The format specifiers are actually positional rather than by name so they're for information only.

PR #68, opened 2024-04-22
Provide MQL logging for first/single  queries.

damieng:


JamesKovacs (approved):
LGTM

Unfortunate that we have to use reflection. We should make `IMongoQueryProvider` public in the next driver release so we don't have to do this juggling.

PR #69, opened 2024-04-23
Support IEnumerable<T> props providing actually T[] or IList<T>

damieng:
Expands support of the property declarations we support to include `IEnumerable<T>` not just `T[]` or `IList<T>` implementations.

Note: The actual concrete type of the class must still support `T[]` or `IList<T>` as these are requirements made by EF for the change tracking system.

Implements EF-111, supersedes #63


JamesKovacs (approved):
LGTM

PR #70, opened 2024-04-24
Clean-up of warnings and best suggestions

damieng:
The smallest code/style format I could manage and be happy with broken out one type per commit for easy review. Almost all of these were automatic except the first (comments only) and fourth.

1. Corrects 3 files copyright notices from 2024 to 2023 for project-wide standard header (manual)
2. Uses "var" instead of declared type and updates .editorconfig to enforce  (auto)
3. Removed some unused usings (auto)
4. Switch from using deprecated DeclaredEntityType (manual)
5. Standardize on the EF style static naming (auto)
6. A few more vars (auto)
7. Use the collection initializer syntax (auto)
8. Switch back to using compiled regex we dropped previously to support EF6 (auto)
9. Null check pattern (auto)
10. Correct a variable name (auto)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -46,9 +46,9 @@ dotnet_style_qualification_for_method = false:suggestion
 dotnet_style_qualification_for_event = false:suggestion
 
 # Types: use keywords instead of BCL types, and permit var only when the type is clear
-csharp_style_var_for_built_in_types = false:suggestion
-csharp_style_var_when_type_is_apparent = false:none
-csharp_style_var_elsewhere = false:suggestion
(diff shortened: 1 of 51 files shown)


rstam:
> 5. Standardize on the EF style static naming (auto)

Let's not even consider this for the C# project. It is very useful for the the naming convention to make it obvious whether a field is static or instance.

damieng:
It's fairly useful here as many of our classes have to mirror the behavior of EF supplied ones and as they evolve with EF ours need to evolve alongside them.

JamesKovacs (approved):
LGTM

PR #71, opened 2024-04-25
Prevent unsupported type discriminators & BsonTimeSpanOptionsAttribute

damieng:
BsonTimeSpanOptionsAttribute was already coded but not registered.

New code to detect type discriminators and throw as they're not supported in next release.


JamesKovacs (approved):
LGTM

PR #72, opened 2024-04-25
EF-79: Add support and tests for ValueConverters

damieng:
There are some limitations to note:

- We can't fully support mapping CLR bool to anything as there's a bug in the Mongo C# Driver query pipeline for bool mapping
- Enum conversions are automatically setup by EF conventions to do the necessary mapping rather than us doing it ourselves
- We can't support nullable enums defaulting to the first enum any more because of the above

Implements EF-79


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ValueConverterTests.cs
```diff
@@ -0,0 +1,533 @@
...
+    public void ObjectId_can_deserialize_from_string()
+    {
+        var expectedId = ObjectId.GenerateNewId();
+        var expected = new IdIsString { _id = expectedId.ToString() };
+        tempDatabase.CreateTemporaryCollection<IdIsString>().InsertOne(expected);
+
+        var collection = GetCollection<IdIsObjectId>();
+        var db = SingleEntityDbContext.Create(collection, ClrObjectIdToMongoString);
+
+        var found = db.Entitites.First();
```
JamesKovacs, line 53: Entitites => Entities (extra "t" in DbSet<T> name)

JamesKovacs (approved):
LGTM. Minor spelling error to correct before merging.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ValueConverterTests.cs
damieng: I'll resolve that in another PR as it affects a whole bunch of tests and predates the PR.

PR #73, opened 2024-04-26
Throw helpful error message if transactions attempted.

damieng:
Previously you would get a dependency injection resolution failure that `IDbContextTransactionManager` could not be resolved if you attempted an EF transaction.

Now it displays a helpful message that the MongoDB EF Core Provider does not support transactions.

The `IDbContextTransactionManager` process also requires an `IDatabaseCreator` for some reason which we also don't yet support so we have a placeholder here too.

We can then adjust this in a future update once we support transactions.


JamesKovacs (approved):
LGTM

PR #74, opened 2024-04-26
Fix Entities typo on SingleEntityDbContext used in tests

damieng:
The test helper class `SingleEntityDbContext` had a typo of `Entitites` instead of `Entities`.

This affects a whole bunch of tests so is an isolated PR.


JamesKovacs (approved):
LGTM

PR #75, opened 2024-04-29
Handle owned entities having owned entities

damieng:
Previously threw an error because the parameter was not correctly remapped into a shaper and replaced.

Now supports serialization, deserialization and querying on owned entities having owned entities.

Fixes EF-83

PR #76, opened 2024-04-29
Add error messaging for untranslatable operations

damieng:
Now throws an `InvalidOperationException` when it detects a query it can't translate. e.g. when using relational bulk operations.

```
The LINQ expression 'DbSet<SimpleEntity>()
    .ExecuteDelete()' could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
```

This doesn't capture every scenario but will capture some of the more common ones and provide better clarity.

Please ignore the first two commits on this PR - it is stacked from #75  - as they will disappear on the rebase with main before merging.

Test failure is the transient EF too many instances error that sometimes happens. Will loop with EF team on any possible mitigations going forward but unrelated to this PR directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# MongoDB Entity Framework Core Provider Preview
+# MongoDB Entity Framework Core Provider 
 
-This project is currently in preview and not recommended for production use yet.
+The MongoDB EF Core Provider requires Entity Framework Core 8 on .NET 8 or later.
 
 ## Getting Started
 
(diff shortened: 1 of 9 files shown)


JamesKovacs (approved):
LGTM

PR #77, opened 2024-04-29
Add some test coverage for global filters

damieng:
Adds a few tests to ensure that EF global filters are supported.

Related to EF-112 but that ticket needs more info for a full resolution.


JamesKovacs (approved):
LGTM

PR #78, opened 2024-04-30
Detect unsupported attribute properties

damieng:
Although we support the Table and Column attribute we don't support Table.Schema or Column.TypeName parts of them.

This PR throws if we encounter those.

We are intentionally allowing the unsupported Column.Order through as it has no data loss implications.


BorisDog (approved):
LGTM

PR #86, opened 2024-05-01
Test clean-up

damieng:
Cleans up the unit and functional tests a little, specifically:

- Always use "db" instead of "dbContext" or "context" 
- Add additional context warning suppression configuration to prevent CI and code coverage warnings
- Dispose of the DbContext in a lot more places


damieng:
Test only changes.

PR #89, opened 2024-05-03
Implement IDatabaseCreator

damieng:
Implements IDatabaseCreator fully rather than throwing NotSupportedExceptions so common sample code like `dbContext.Database.EnsureCreated()` executes without error (even though it's not required on MongoDB as it will create on first document insertion)

Fixes EF-114


JamesKovacs (approved):
LGTM

PR #91, opened 2024-05-10
Fix exception caused by sequential single/first calls

damieng:
A late change in 8.0.0 included a workaround for logging first/single calls which uses reflection.

Unfortunately the caching mechanism fails on subsequent calls when the underlying query type changes and the MongoClientWrapper is being reused.

This stops trying to cache the reflection call. A newer version of the Mongo C# Driver will remove the need to reflect entirely.

Also took the opportunity to increase test coverage.

Fixes EF-120


sanych-sun (approved):
LGTM

PR #92, opened 2024-05-13
Support lists of lists for primitive types

damieng:
We supported lists of lists of primitives in preview release 1 as we had custom comparer code. Moving to EF8 caused this to stop as the built-in comparers did not support it.

This adds support again by back-porting the EF Core 9 comparers to our provider.  Non-primitives are not supported directly in lists of lists due to limitations of EF Core 8 & 9 and instead you need to ensure an intermediate owned entity type between them instead.

Fixes EF-116


sanych-sun (approved):
LGTM

PR #93, opened 2024-05-13
Fix Api Docs script for Patch Release

sanych-sun:


damieng:
After discussion with the EF team they have suggested we only produce patch releases as not to confuse MongoDB EF Core users into thinking our 8.1.0 requires an EF 8.1 release.

PR #94, opened 2024-05-17
Add Decimal128 conversion

damieng:
Add a default conversion for Decimal <> Decimal128 conversions as a precursor for the BSON Representation attribute work.

Fixes EF-125


sanych-sun (approved):
LGTM

PR #95, opened 2024-05-18
Use actual DateTimeKind specified not just local

damieng:
Problem was masked by precision reduction code in tests.

Fixes EF-126


JamesKovacs (approved):
LGTM

PR #96, opened 2024-05-21
Use Mongo C# Driver for enum mapping over built-in EF converters

damieng:
The built-in converters don't handle null internally (we have a separate issue to tackle that for non-enum value converters)

This changes Enum handling to go back to our built-in conversion which will also retain better compatibility with the C# driver.

Fixes EF-128


JamesKovacs (approved):
LGTM

PR #97, opened 2024-05-23
BsonType configuration 

damieng:
Provides a `HasBsonType` fluent API that allows the model builder to configure specific properties to be stored as specific BSON types.

Fixes EF-127

A subsequent PR will further expose this feature out so that `BsonRepresentation` attributes can be recognized (EF-44)

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
+++ b/src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
@@ -92,13 +92,13 @@ public static class MongoPropertyBuilderExtensions
     /// Configures the <see cref="BsonType"/> that the property is stored as when targeting MongoDB.
     /// </summary>
     /// <param name="propertyBuilder">The builder for the property being configured.</param>
-    /// <param name="bsonType">The <see cref="BsonType"/> to store this property as.</param>
+    /// <param name="bsonType">The <see cref="BsonType"/> to store this property as
+    /// or <see langword="null" /> to unset the configuration and use the default.</param>
     /// <returns>The same builder instance so that multiple calls can be chained.</returns>
     public static PropertyBuilder HasBsonType(
         this PropertyBuilder propertyBuilder,
-        BsonType bsonType)
+        BsonType? bsonType)
     {
-        ArgumentNullException.ThrowIfNull(bsonType);
         propertyBuilder.Metadata.SetBsonType(bsonType);
         return propertyBuilder;
     }
@@ -108,11 +108,12 @@ public static class MongoPropertyBuilderExtensions
     /// </summary>
     /// <typeparam name="TProperty">The type of the property being configured.</typeparam>
     /// <param name="propertyBuilder">The builder for the property being configured.</param>
-    /// <param name="bsonType">The <see cref="BsonType"/> to store this property as.</param>
+    /// <param name="bsonType">The <see cref="BsonType"/> to store this property as
+    /// or <see langword="null" /> to unset the configuration and use the default.</param>
     /// <returns>The same builder instance so that multiple calls can be chained.</returns>
     public static PropertyBuilder<TProperty> HasBsonType<TProperty>(
         this PropertyBuilder<TProperty> propertyBuilder,
-        BsonType bsonType)
+        BsonType? bsonType)
         => (PropertyBuilder<TProperty>)HasBsonType((PropertyBuilder)propertyBuilder, bsonType);
 
     /// <summary>
--- a/src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
+++ b/src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
@@ -1,20 +1,21 @@
 ﻿/* Copyright 2023-present MongoDB Inc.
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with the License.
-* You may obtain a copy of the License at
-*
-* http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing, software
-* distributed under the License is distributed on an "AS IS" BASIS,
-* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-* See the License for the specific language governing permissions and
-* limitations under the License.
-*/
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
 using System;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using MongoDB.Bson;
 using MongoDB.EntityFrameworkCore.Metadata;
 
 // ReSharper disable once CheckNamespace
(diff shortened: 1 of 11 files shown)


src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
```diff
@@ -53,65 +54,127 @@ public static class MongoPropertyBuilderExtensions
         => (PropertyBuilder<TProperty>)HasElementName((PropertyBuilder)propertyBuilder, name);
 
     /// <summary>
-    /// Configures the <see cref="DateTimeKind"/> for the property.
+    /// Configures the document element that the property is mapped to when targeting MongoDB.
+    /// If an empty string is supplied then the property will not be persisted.
```
JamesKovacs, line 58: Although not recommended, an empty string is a valid field name in MongoDB. `null` however is not a valid field name. Is an empty string not mapping a property an EF convention or something that we are introducing?

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
```diff
@@ -53,65 +54,127 @@ public static class MongoPropertyBuilderExtensions
...
+    /// <returns>The same builder instance if the configuration was applied, <see langword="null" /> otherwise.</returns>
+    public static IConventionPropertyBuilder? HasElementName(
+        this IConventionPropertyBuilder propertyBuilder,
+        string? name,
+        bool fromDataAnnotation = false)
     {
-        propertyBuilder.Metadata.SetDateTimeKind(dateTimeKind);
+        if (!CanSetElementName(propertyBuilder, name, fromDataAnnotation))
+        {
+            return null;
```
JamesKovacs, line 71: Should we not be throwing an `InvalidOperationException` in this case? If we return `null`, then we will throw a `NullReferenceException` when we chain the fluent API.

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
```diff
@@ -53,65 +54,127 @@ public static class MongoPropertyBuilderExtensions
...
+    /// </summary>
+    /// <param name="propertyBuilder">The builder for the property being configured.</param>
+    /// <param name="bsonType">The <see cref="BsonType"/> to store this property as.</param>
+    /// <returns>The same builder instance so that multiple calls can be chained.</returns>
+    public static PropertyBuilder HasBsonType(
+        this PropertyBuilder propertyBuilder,
+        BsonType bsonType)
     {
-        propertyBuilder.Metadata.SetDateTimeKind(dateTimeKind);
+        ArgumentNullException.ThrowIfNull(bsonType);
```
JamesKovacs, line 101: `BsonType` is an enum and cannot be null.

JamesKovacs: Do we want to make this `BsonType?` so that `null` is valid and can be used to unset a previous `HasBsonType` configuration?

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
```diff
@@ -103,19 +113,26 @@ private static IBsonSerializer CreateTypeSerializer(Type type, IReadOnlyProperty
...
-            _ when type == typeof(Decimal128) => new Decimal128Serializer(),
             _ when type.IsEnum => EnumSerializer.Create(type),
+
+            // Nullable
             {IsGenericType: true} when type.GetGenericTypeDefinition() == typeof(Nullable<>)
                 => CreateNullableSerializer(type.GetGenericArguments()[0]),
+
+            // Mongo structs
+            _ when type == typeof(ObjectId) => ObjectIdSerializer.Instance,
+            _ when type == typeof(Decimal128) => new Decimal128Serializer(),
```
JamesKovacs, line 131: We should configure serializers for ObjectId and Decimal128, but this seems unrelated to the PR at hand. Move to a separate PR?

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Exerciser.cs
JamesKovacs, line 1: This seems related to adding support for `ObjectIdSerializer`, not `HasBsonType` support.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Utilities/ExtensionMethods.cs
```diff
@@ -36,9 +37,12 @@ public static byte NextByte(this Random random)
...
-        =>  dateTime.ToString("yyyy-MM-ddTHH:mm:ss.fffK");
-
     public static void WriteTestDocs<T>(this IMongoCollection<T> collection, IEnumerable<T> docs)
         => collection.BulkWrite(docs.Select(p => new InsertOneModel<T>(p)));
+
+    public static DateTime ToBsonPrecision(this DateTime dateTime)
+    {
+        var bsonDateTime = new BsonDateTime(dateTime);
+        return dateTime.Kind == DateTimeKind.Utc ? bsonDateTime.ToUniversalTime() : bsonDateTime.ToLocalTime();
+    }
```
JamesKovacs, line 47: Unrelated to `HasBsonType` support. Move to separate PR.

JamesKovacs (changes requested):
`HasBsonType` looks good but the unrelated work should be moved into their own PRs.

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
damieng: This is the pattern used across all EF providers. You can see it in the Cosmos, SqlServer and relational base providers.

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
damieng: This was just improving the documentation. EF uses this convention in other providers. If we want to support writing to an element with an empty name we may be able to that but we'd need to create a new ticket and figure out an alternative marker for unmapped properties.

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
damieng: Yeah, let's do that.

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
damieng: This doesn't add/remove any support, it just groups them together with comments as the list was getting long. I can revert.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Exerciser.cs
damieng: Just a minor reordering while I was in there. No new code.

damieng: This is a new class for testing round-tripping data back and forth to MongoDB with an extra check via the C# driver. It came out of writing all the new BsonType tests as they were getting repetitive. It's not related to any other work.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Utilities/ExtensionMethods.cs
damieng: This is test-only code and relates to testing DateTime's which I need for the new BsonType tests that touch BsonDateTime. 

I had a ToExpectedPrecision method I was previously using for some other BsonDateTime related tests but it didn't handle UTC. The new method is used for both the new tests and the old.

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyBuilderExtensions.cs
damieng: Now done.

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
damieng: Removed this clean-up from the PR.

JamesKovacs (approved):
LGTM

PR #98, opened 2024-05-24
EF-131: Support matching on owned entity values

damieng:
Previously we did not support matching entire owned entities for where/contains/first etc. 

This adds support for that by ensuring the entity serializer can serialize the EF entity back to the Bson document for matching.

Fixes EF-131.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
+++ b/src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
@@ -14,11 +14,9 @@
  */
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Metadata;
-using MongoDB.Bson.IO;
 using MongoDB.Bson.Serialization;
 using MongoDB.EntityFrameworkCore.Extensions;
 
@@ -60,34 +58,22 @@ namespace MongoDB.EntityFrameworkCore.Serializers
                 return;
             }
 
-            var storedKeyProperties = GetStoredKeyProperties();
-            if (storedKeyProperties.Any())
-            {
-                // Has key, throw and indicate user should compare by key
-                var keys = string.Join(", ", storedKeyProperties.Select(p => "'" + p.Name + "'"));
-                throw new NotSupportedException($"Entity to entity comparison is not supported when entities have keys. Compare '{_entityType.DisplayName()}' entities by {keys} instead.");
-            }
+            // We do not support direct entity serialization right now because:
+            //  - Matching owned entities by example or expression may mismatch because of unmapped fields/default values
+            //    (we will likely configurable policy for this)
+            //  - Matching root/entities with keys should be done by key fields, not by entity comparison
+            //    (we will rewrite the expression tree to do this automatically in a future update)
 
-            // No key, no identity, so compare by value
-            context.Writer.WriteStartDocument();
-            WriteProperties(context.Writer, value, _entityType.GetProperties().Where(p => _isStored(p)));
-            context.Writer.WriteEndDocument();
+            var storedKeyProperties = GetStoredKeyProperties();
+            var uniqueness = storedKeyProperties.Any()
+                ? string.Join(", ", storedKeyProperties.Select(p => "'" + p.Name + "'"))
+                : "unique fields";
+
+            throw new NotSupportedException($"Entity to entity comparison is not supported. Compare '{_entityType.DisplayName()}' entities by {uniqueness} instead.");
         }
 
         private IReadOnlyProperty[] GetStoredKeyProperties()
             => _entityType.FindPrimaryKey()?.Properties.Where(_isStored).ToArray() ?? [];
-
-        private static void WriteProperties(IBsonWriter writer, TValue entry, IEnumerable<IReadOnlyProperty> properties)
-        {
-            foreach (var property in properties)
-            {
-                var serializationInfo = SerializationHelper.GetPropertySerializationInfo(property);
-                var elementName = serializationInfo.ElementPath?.Last() ?? serializationInfo.ElementName;
-                writer.WriteName(elementName);
-                var context = BsonSerializationContext.CreateRoot(writer);
-                serializationInfo.Serializer.Serialize(context, property.PropertyInfo.GetValue(entry));
-            }
-        }
 
         void IBsonSerializer.Serialize(BsonSerializationContext context, BsonSerializationArgs args, object value)
             => Serialize(context, args, (TValue)value);
--- a/src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
+++ b/src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
@@ -1,20 +1,24 @@
 ﻿/* Copyright 2023-present MongoDB Inc.
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with the License.
-* You may obtain a copy of the License at
-*
-* http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing, software
-* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
```diff
@@ -48,7 +53,49 @@ public TValue Deserialize(BsonDeserializationContext context, BsonDeserializatio
...
-            => throw new NotImplementedException();
+        {
+            if (value == null)
+            {
+                context.Writer.WriteNull();
+                return;
+            }
+
+            context.Writer.WriteStartDocument();
+            WriteKey(context.Writer, value);
```
rstam, line 64: Does every class have a key?

Do we not support compound types that don't have a key, like address?

src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
```diff
@@ -48,7 +53,49 @@ public TValue Deserialize(BsonDeserializationContext context, BsonDeserializatio
...
+        }
+
+        private static void WriteProperties(IBsonWriter writer, TValue entry, IEnumerable<IReadOnlyProperty> properties)
+        {
+            foreach (var property in properties)
+            {
+                var serializationInfo = SerializationHelper.GetPropertySerializationInfo(property);
+                var elementName = serializationInfo.ElementPath?.Last() ?? serializationInfo.ElementName;
+                writer.WriteName(elementName);
+                var root = BsonSerializationContext.CreateRoot(writer);
```
rstam, line 95: We usually call this variable context but you're free to keep whatever name you want.

rstam (commented):
Looks good. I have a few questions.

src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
damieng: Will rename.

src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
damieng: So this comment led to a thought which led to a discussion with the EF team on identity and what it means.

The preference from the EF world is that root entities have identity and thus should be compared only by identity for a variety of reasons.

Meanwhile entities without keys (such as owned entities by default) have no identity (key) and so should be compared by value.

I think we need to dig more into the comparison of key based entities as if we go down the EF route we'll have to rewrite the tree to change entity comparisons to be by key rather than serializing the entity. If we decide not to go down that route but to align more with the existing C# LINQ driver then we'll need to understand and document the implications.

In the mean-time we'll throw for that case so we can proceed with null comparisons (the original issue) and keyless by-value comparisons (no debate).

damieng:
Seeing some failures when targeting older servers. Can you provide any insight @rstam ?

rstam:
> Seeing some failures when targeting older servers. Can you provide any insight @rstam ?

Can you point me to something specific you are asking about?

rstam:
To intercept the pipeline that is causing the error I think you could use the logging hooks you have already added right?

damieng:
As it throws and only happens on earlier versions of MongoDB it wasn't quite as easy to get the MQL.

```EFCoreTest-2024-06-01T08-14-26-1.planets.aggregate([{ "$match" : { "$expr" : { "$eq" : ["$$ROOT", null] } } }])```

I'm able to reproduce this directly with the C# Driver with a simple Where query comparing the entity to null - either directly (unlikely to ever be written) or against a variable which happens to contain null (much more likely). I'll raise an issue on Jira.

**Comment by rstam** [comment]
> EFCoreTest-2024-06-01T08-14-26-1.planets.aggregate([{ "$match" : { "$expr" : { "$eq" : ["$$ROOT", null] } } }])

Huh? 

The root document can NEVER be null.

What was the C# that generated this query? Can you detect that as an invalid query before passing it to the driver?

File: tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/OwnedEntityTests.cs [hunk 3]
```diff
@@ -41,6 +41,103 @@ public void OwnedEntity_nested_one_level_materializes_single()
         Assert.Equal(Location1.longitude, actual.location.longitude);
     }
 
+    [Fact]
+    public void OwnedEntity_nested_one_level_where_null()
```
sanych-sun, line 45: This and next test have confusing names. `where_null`, but we actually here we filter by != null. When next test named as `where_not_null` and there we filtering by == null.

src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
```diff
@@ -48,7 +53,41 @@ public TValue Deserialize(BsonDeserializationContext context, BsonDeserializatio
...
+            if (value == null)
+            {
+                context.Writer.WriteNull();
+                return;
+            }
+
+            var storedKeyProperties = GetStoredKeyProperties();
+            if (storedKeyProperties.Any())
+            {
+                // Has key, throw and indicate user should compare by key
```
sanych-sun, line 66: Is there any better place for the code that is clearly related to entities comparison only? It's confusing why EntitySerializer trying to validate  comparison operation inside of Serialize method.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/WhereTests.cs
```diff
@@ -251,6 +251,39 @@ public void Where_string_list_any()
...
+    {
+        var actual = _db.Planets.First(p => p != null);
+        Assert.NotNull(actual);
+    }
+
+    [Fact]
+    public void Where_entity_equal_throws()
+    {
+        var first = _db.Planets.First();
+        var ex = Assert.Throws<NotSupportedException>(() => _db.Planets.Where(p => p == first).ToList());
```
sanych-sun, line 272: Could you please elaborate why we throws here? It looks like legitimate way to use for me.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/WhereTests.cs
```diff
@@ -251,6 +251,39 @@ public void Where_string_list_any()
...
+        Assert.Contains(nameof(Planet._id), ex.Message);
+        Assert.Contains(nameof(Planet), ex.Message);
+    }
+
+    [Fact]
+    public void Where_entity_not_equal_throws()
+    {
+        var orderedPlanets = _db.Planets.OrderBy(p => p.name);
+        var first = orderedPlanets.First();
+        var ex = Assert.Throws<NotSupportedException>(() => orderedPlanets.First(p => p != first));
```
sanych-sun, line 282: Same as above.

sanych-sun (changes requested):

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/WhereTests.cs
damieng: There's some debate/discussion as to what this should do.

The EF team's preference is to rewrite this as a id based comparison for items with keys and a full object comparison for items that don't have identity.

Robert has some reservations about that and it would mean that the C# Driver and the EF Provider diverge on behavior here.

If an item has changed in-memory do you expect to be able to get the original one still? With the EF method we would but with the C# Driver we won't.

The original bug this PR is addressing was really just around null and so I want to get this unblocked and shipped and then we can move on to discussion and resolution of actual entity comparison next release when had more time to discuss.

src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
damieng: This code would be removed once the comparison issue is solved next release.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/OwnedEntityTests.cs
damieng: Nice catch, fix pushed.

sanych-sun (approved):
LGTM

PR #101, opened 2024-06-01
EF-142: Log MQL even when LINQ driver throws.

damieng:
Previously if the LINQ provider threw the MQL would not be logged.

This change ensures we always log the MQL even if it does throw as it is very useful in debugging certain throwing scenarios.

Small test cleanup in the touched LoggingTests to reduce boilerplate in finding the correct log entry.

Fixes EF-142


BorisDog (approved):
LGTM

PR #103, opened 2024-06-07
Address minor code issues identified in Coverity

damieng:
Nothing major, mostly resource allocation in tests plus some null checks in the library.

### DIFF
--- a/tests/MongoDB.EntityFrameworkCore.FunctionalTests/Storage/MongoDatabaseCreatorTests.cs
+++ b/tests/MongoDB.EntityFrameworkCore.FunctionalTests/Storage/MongoDatabaseCreatorTests.cs
@@ -76,7 +76,7 @@ public class MongoDatabaseCreatorTests
     [Fact]
     public void EnsureDeleted_returns_false_when_it_does_not_exist()
     {
-        var database = new TemporaryDatabaseFixture();
+        using var database = new TemporaryDatabaseFixture();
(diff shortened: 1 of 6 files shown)


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Storage/MongoDatabaseCreatorTests.cs
```diff
@@ -76,7 +76,7 @@ public void EnsureDeleted_returns_true_and_deletes_database_when_it_exists()
     [Fact]
     public void EnsureDeleted_returns_false_when_it_does_not_exist()
     {
-        var database = new TemporaryDatabaseFixture();
+        using var database = new TemporaryDatabaseFixture();
```
sanych-sun, line 79: Should we inject the fixture instead?

sanych-sun (changes requested):

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Storage/MongoDatabaseCreatorTests.cs
damieng: No, this test deletes the database. That would break the other tests.

damieng: Just keeping the database creation/deletion tests isolated to avoid concurrency issues.

PR #104, opened 2024-06-10
EF-144: Add support for Compiled Models

damieng:
Adds support for the Compiled Model feature of Entity Framework Core.

Fixes EF-144.


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Design/CompiledModelTests.cs
```diff
@@ -0,0 +1,177 @@
...
+        Assert.Single(files, f => f.Path == nameof(SimpleContext) + "ModelBuilder.cs");
+        Assert.Single(files, f => f.Path == nameof(EveryType) + "EntityType.cs");
+        Assert.Single(files, f => f.Path == nameof(OwnedEntity) + "EntityType.cs");
+    }
+
+    [Fact]
+    public void Regenerate_model_code_for_roundtrip_tests()
+    {
+        // This will re-generate local code-generated model files for SimpleContext
+        // If they change you should run the tests again and check them in.
```
sanych-sun, line 83: Can we try to find more elegant way to re-generate the code? It might be an option to generate the code on fly and load the generated assembly dynamically, so we do not even need to check-in the generated code.

sanych-sun (changes requested):

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Design/CompiledModelTests.cs
damieng: I could look into that in the future but I think it's a lot lower pri than a lot of stuff on the backlog right now. I'll raise a ticket and put it in there.

damieng: EF9 and .NET 9 actually have a build process that does this automatically so I think we should just use that in due course.

JamesKovacs: Given that EF9 solves this exact problem in a standard way, I agree with the approach of filing an EF ticket to track this work in EF9 and checking in the generated test files until that work is completed.

JamesKovacs (approved):
LGTM

PR #108, opened 2024-06-13
Upgrade to C# Driver 2.26.0

damieng:
As well as updating the NuGet reference there are two small changes that directly relate to 2.26.0:

1. The unit test that previously failed because of a bug in the C# Driver is now re-enabled
2. The logging code no longer needs to use reflection and is cleaned up to reflect the simpler pattern


sanych-sun (approved):
LGTM

PR #109, opened 2024-06-13
EF-44: Bson Representation support

damieng:
Adds support for BsonRepresentationAttribute which also required we re-work the existing (unpublished) HasBsonType into HasBsonRepresentation so it can take more than just a type but also overflow and truncation properties.

First two commits are required but from another branch - will rebase once they are merged.


src/MongoDB.EntityFrameworkCore/Metadata/BsonRepresentationConfiguration.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Collections.Generic;
+using MongoDB.Bson;
+
+namespace MongoDB.EntityFrameworkCore.Metadata;
+
+/// <summary>
+/// Configuration of a BSON representation defining how the value from property
+/// should be stored in the MongoDB database.
+/// </summary>
+public record BsonRepresentationConfiguration
```
sanych-sun, line 26: public readonly record?

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
```diff
@@ -88,17 +90,30 @@ private static IBsonSerializer CreateTypeSerializer(IReadOnlyProperty property)
 
         var typeSerializer = CreateTypeSerializer(property.ClrType, property);
 
-        // Apply HasBsonType configuration if set
-        var bsonType = property.GetBsonType();
-        if (bsonType != null && typeSerializer is IRepresentationConfigurable representationConfigurable)
+        return property.GetBsonRepresentation() is { } bsonRepresentation
```
sanych-sun, line 93: Personally I do not like this weird way to check for not null value. IMHO, the following code doing the same but easier to read:
`var bsonRepresentation = property.GetBsonRepresentation();`
`return bsonRepresentation == null ? typeSerializer : ApplyBsonRepresentation(bsonRepresentation, typeSerializer);`

sanych-sun (changes requested):

src/MongoDB.EntityFrameworkCore/Metadata/BsonRepresentationConfiguration.cs
damieng: public readonly record struct?

damieng: Can only use readonly if we change it to struct. Given that this is an optional/nullable in a bunch of places that means you won't be able to access .BsonType without going through .Value everywhere.

e.g. in the tests

```csharp
Assert.NotNull(representation);
Assert.Equal(bsonType, representation.BsonType);
Assert.Equal(allowOverflow, representation.AllowOverflow);
Assert.Equal(allowTruncation, representation.AllowTruncation);
```

Becomes

```csharp
Assert.NotNull(representation);
Assert.Equal(bsonType, representation.Value.BsonType);
Assert.Equal(allowOverflow, representation.Value.AllowOverflow);
Assert.Equal(allowTruncation, representation.Value.AllowTruncation);
```

Which I don't think is an improvement.

src/MongoDB.EntityFrameworkCore/Metadata/BsonRepresentationConfiguration.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    /// <summary>
+    /// Create a <see cref="BsonRepresentationConfiguration"/> from an IDictionary representation
+    /// so it may be recreated from an EF Core static model.
+    /// </summary>
+    /// <returns>A <see cref="BsonRepresentationConfiguration"/> with the properties set from the dictionary.</returns>
+    /// <remarks>The dictionary passes is typically created by <see cref="ToDictionary"/>.</remarks>
+    public static BsonRepresentationConfiguration CreateFrom(IDictionary<string, object> values)
+    {
+        var bsonType = (BsonType?)values["BsonType"] ?? throw new InvalidOperationException("BsonType is required.");
```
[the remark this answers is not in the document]
damieng: Updated to use TryGetValue and throw if not found.

src/MongoDB.EntityFrameworkCore/Serializers/SerializationHelper.cs
damieng: I think I prefer the pattern matching version. Yes, in isolation it's a bit unusual but once you start using pattern matching everywhere it fits well with that.

sanych-sun (approved):
LGTM

PR #112, opened 2024-06-19
Add test to ensure partial updates of roots

damieng:
Just a simple test to ensure we are only updating fields that have changed at a root level.

Partial updates beyond the root level are being considered in EF-147.


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/UpdateEntityTests.cs
```diff
@@ -48,6 +48,19 @@ class RealisticEntity
         public DateTime lastModified { get; set; }
     }
 
+    class OwningEntity
```
BorisDog, line 51: Is this used somewhere?

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/UpdateEntityTests.cs
```diff
@@ -106,6 +120,52 @@ public void Update_realistic_entity()
...
+
+        // Cause two updates to happen interleaved
+        {
+            using var db1 = SingleEntityDbContext.Create(collection);
+            var entity1 = db1.Entities.First();
+            entity1.lastCount++;
+
+            using var db2 = SingleEntityDbContext.Create(collection);
+            var entity2 = db2.Entities.First();
+            entity2.session = session2;
```
BorisDog, line 150: Is same value assignment considered as modification (`entity2.session = entity2.session`)?
Whatever the answer is, is it important to capture this behavior in test?

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/UpdateEntityTests.cs
damieng: Ah, that was used in a different test that I decided not to add, now removed.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Update/UpdateEntityTests.cs
damieng: This creates two copies of the same entity and sets the properties to different values. There's no same-assignment path here.

BorisDog: Yup, the question is whether the same value assignment is considered modification, and whether we should capture the existing behavior in a testcase. For example:

```
entity1.lastCount++;
entity1.session = entity1.session;

entity2.session = session2;

db2.SaveChanges();
db1.SaveChanges(); // Will this overwrite session to  entity1.session ?
```

damieng: EF shouldn't consider that a change so we don't have test coverage for it as it's out of scope for what our driver actually sees.

BorisDog (approved):
LGTM.
Notice that tests are failing.

PR #117, opened 2024-07-02
EF-47: Dictionary & ReadOnlyDictionary property support

damieng:
Implements EF-47 for Dictionary/ReadOnlyDictionary support.

Note: One test is disabled as there is a bug in the C# Driver that prevents this indexers being used on IReadOnlyDictionary. Being tracked as CSHARP-5171,

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Serializers/DictionarySerializationProvider.cs
+++ b/src/MongoDB.EntityFrameworkCore/Serializers/DictionarySerializationProvider.cs
@@ -67,7 +67,7 @@ internal class DictionarySerializationProvider : BsonSerializationProviderBase
             return CreateGenericSerializer(typeof(DictionaryInterfaceImplementerSerializer<,,>), [concreteType, keyType, valueType], serializerRegistry);
         }
 
-        var readOnlyDictionaryInterface= genericTypeDefinition == typeof(IReadOnlyDictionary<,>)
+        var readOnlyDictionaryInterface = genericTypeDefinition == typeof(IReadOnlyDictionary<,>)
             ? type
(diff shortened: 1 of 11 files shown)


src/MongoDB.EntityFrameworkCore/Serializers/DictionarySerializationProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+    public static bool Supports(Type type)
+        => type.IsGenericType && SupportedDictionaryTypes.Contains(type.GetGenericTypeDefinition());
+
+    public override IBsonSerializer GetSerializer(Type type, IBsonSerializerRegistry serializerRegistry)
+    {
+        ArgumentNullException.ThrowIfNull(type);
+
+        if (type is {IsGenericType: true, ContainsGenericParameters: true})
+        {
+            throw new ArgumentException($"Generic type '{type.ShortDisplayName()}' has unassigned generic type parameters.",
```
sanych-sun, line 43: Should it says "Open generic types are not supported"?

sanych-sun (approved):
LGTM + some suggestions.

src/MongoDB.EntityFrameworkCore/Serializers/DictionarySerializationProvider.cs
damieng: I went with this naming as the docs page for ContainsGenericParameters at https://learn.microsoft.com/en-us/dotnet/api/system.reflection.methodbase.containsgenericparameters?view=net-8.0 refers to them as

> true if the current MethodBase object represents a generic method that contains unassigned generic type parameters; otherwise, false.

PR #118, opened 2024-07-04
EF-81 & EF-150: SaveChanges transactions & ConcurrencyCheck support

damieng:
Fixes EF-81: Implement transaction support in SaveChanges(Async)
Fixes EF-150: Add support for ConcurrencyCheck attribute


damieng:
Normalizing and splitting up for easier review.

PR #121, opened 2024-07-23
Upgrade EF to 8.0.7 and MongoDB.Driver to 2.28.0

damieng:
Had to remove an array test as 2.28.0 doesn't support mixed-value arrays any more - CSHARP-4957


sanych-sun (approved):
LGTM

PR #125, opened 2024-07-29
Implement RowVersion and Concurrency check

damieng:
Implements EF-150. 

This branch includes the prior EF-81 work (first commit). That should not be part of the review but is required here for tests to run. I will rebase it before merge to remove it once EF-150 is in.


adelinowona (approved):
LGTM!

PR #126, opened 2024-07-30
Do not delete DB after each fixture

damieng:
Testing to isolate issues with transactions on CI.


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Utilities/TemporaryDatabaseFixture.cs
```diff
@@ -93,8 +93,10 @@ public IMongoCollection<T> GetExistingTemporaryCollection<T>([CallerMemberName]
 
 
     public void Dispose()
-        => Client.DropDatabase(MongoDatabase.DatabaseNamespace.DatabaseName);
+    {
```
sanych-sun, line 96: What the problem with dropping database? I do remember you said it somehow conflicts with transactions, but why/how? Clean up should be happening when all tests are done and transactions are completed.

PR #127, opened 2024-08-01
Track state of MongoTransaction and ensure consistent use.

damieng:
Just additional check logic to ensure consistency.


janiavdv (approved):
LGTM, just check the failing EG test.

damieng:
Exception is unrelated, will investigate separately.

PR #129, opened 2024-08-05
EF-154: Switch Guid binary format over to V3/Standard

damieng:
Originally considered exposing an API to allow and specify how the Guid was stored but with the move to V3/Standard and the other formats effectively being deprecated it makes sense just to change the defaults.

This will be a breaking change and will need to be called out in the release notes.


adelinowona (approved):
LGTM

PR #131, opened 2024-08-07
EF-155: Create empty collections when inside CreateDatabase

damieng:
CreateDatabase(Async) was previously a no-op as we allowed the database and collections to create on-demand.

Now that transactions are enabled by default there are some situations where creating a collection or database inside the transaction can cause issues.

Instead now we will create any missing empty collections from inside CreateDatabase to avoid these situations as it should be called at app start-up outside of any transaction.


adelinowona (approved):
LGTM. 

Also I see some tests are failing but seems it is still those other tests that failed with the `Unable to read from a snapshot due to pending collection catalog changes` message that we talked about in the transaction PRs.

PR #135, opened 2024-08-14
Clean up some conventions

damieng:
Minor clean-up to some conventions for performance and ease of use.

Aware of small breaks, will add to the docs for 8.1.0


adelinowona (approved):
LGTM

PR #142, opened 2024-08-27
Ensure EF metadata used for querying owned items in collections

damieng:
Includes some refactoring to simplify the creation of the various serializers into BsonSerializerFactory which replaced EntitySerializerCache and eliminates SerializationHelper.

Fixes original bug EF-156 but turns out that it was not limited to just enums nor just type conversions - the query pipeline was not using any EF metadata for the collections of owned entities (but serializtion and deserialization were fine).


adelinowona (approved):
LGTM! + plus I see you disabled tests parallelization; why is that?

damieng:
Accidentally checked that file in - I'd re-enabled test parallization as it works fine locally and forgot to uncomment it again. Now fixed.

papafe (approved):
LGTM for what I can understand 😁

PR #146, opened 2024-09-09
EF-163: Set V3 Guid mode at initialization

damieng:
Previously the V3 Guid mode was set when a query was run. This means if data was inserted before any queries it used the V2 non-standard mode resulting in exception messages.

This moves the configuration of the BsonDefaults static to DbContext initialization.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Infrastructure/MongoModelRuntimeInitializer.cs
+++ b/src/MongoDB.EntityFrameworkCore/Infrastructure/MongoModelRuntimeInitializer.cs
@@ -1,0 +1,66 @@
+﻿/* Copyright 2023-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


jakoss (approved):

src/MongoDB.EntityFrameworkCore/Infrastructure/MongoModelRuntimeInitializer.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace MongoDB.EntityFrameworkCore.Infrastructure;
+
+/// <summary>
+/// Initializes a <see cref="IModel" /> with the runtime dependencies.
+/// </summary>
+/// <param name="dependencies"></param>
+public class MongoModelRuntimeInitializer(ModelRuntimeInitializerDependencies dependencies)
+    : ModelRuntimeInitializer(dependencies)
+{
+    private static readonly Dictionary<Type, IDiscriminatorConvention>? DiscriminatorConventionDictionary =
```
adelinowona, line 36: This doesn't seem to be used right now. Is it needed?

adelinowona (approved):
LGTM just one comment

src/MongoDB.EntityFrameworkCore/Infrastructure/MongoModelRuntimeInitializer.cs
damieng: Oops! This PR took the code from the upcoming type disciminator PR but I accidentally left that in. Now removed. Thanks!

PR #148, opened 2024-09-17
Add project config to allow 2.28 or 3.0 driver

damieng:
Give us a project configuration "Debug 3.0 Driver" which allows using the MongoDB C# Driver 3.0 pre-release.

This will help us continue developing and provide an easy migration when we switch to the 3.0 driver after it is released.


alrz:
3.0 has been released for a few days, is the update scheduled for next major release or we may get it earlier?

damieng:
Aiming for early next week.

PR #149, opened 2024-09-19
Fix test that was broken on latest C# Driver branch

damieng:
Latest changes to C# Driver store Decimals as Decimal128.

This exposed a miconfigured test that was supposed to store Decimals as Strings explicitly but was in fact doing it implicitly due to the default value converter configuring the wrong entity.


papafe (approved):
LGTM

PR #154, opened 2024-10-05
Optimize ToTitleCase Method for performance & allocation

wioniqle-q:
Hi,

This pull request introduces optimizations to the `ToTitleCase` method in the `NamingHelperMethods` class to improve performance and reduce memory allocations.

Testing:
-  Passed your unit tests

Key Enhancements:

1. Performance Optimization:
   - Small inputs: 25.4 times faster (4,252.0 ns → 167.3 ns)
   - Medium inputs: 7.3 times faster (105,182.6 ns → 14,419.9 ns)
   - Large inputs: 5.4 times faster (10,983,153.6 ns → 2,048,688.3 ns)
   - Very large inputs: 6.3 times faster (128,103,886.7 ns → 20,264,420.3 ns)

2. Memory Allocation Optimization:
   - Small inputs: 196 times less memory used (9,416 B → 48 B)
   - Medium inputs: 23 times less memory used (45,840 B → 1,992 B)
   - Large inputs: 20 times less memory used (3,926,474 B → 196,977 B)
   - Very large inputs: 18.8 times less memory used (37,083,832 B → 1,968,249 B)

3. Garbage Collection Efficiency:
   - Significant reduction in Gen0, Gen1, and Gen2 collections across all input sizes

performance_test

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/NamingHelperMethods.cs
+++ b/src/MongoDB.EntityFrameworkCore/NamingHelperMethods.cs
@@ -13,6 +13,7 @@
  * limitations under the License.
  */
 
+using System.Buffers;
 using System.Globalization;
 using System.Text.RegularExpressions;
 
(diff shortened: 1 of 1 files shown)


src/MongoDB.EntityFrameworkCore/NamingHelperMethods.cs
```diff
@@ -72,29 +73,59 @@ public static string ToCamelCase(
...
+    public static string ToTitleCase(this string input, CultureInfo culture)
     {
         if (string.IsNullOrEmpty(input)) return input;
 
-        var words = WordSplitRegex().Split(input);
+        Span<char> initialBuffer = stackalloc char[512];
 
-        var result = new char[input.Length];
-        var outIndex = 0;
+        if (input.Length > initialBuffer.Length) return ToTitleCaseLarge(input, culture);
```
rpallares, line 82: You Can probably introduce a const for your small buffer size and allocate the span only if you actually use it.
Here for large input you allocate the small buffer unecesseraly

wioniqle-q: Thank you for your suggestion, but I believe there's a misunderstanding here. The code is already optimized for both small and large inputs:

1. For small inputs (<=512 characters), we use stackalloc which is very efficient as it allocates memory on the stack.
2. For large inputs (>512 characters), we switch to using ArrayPool to rent memory from the heap.

The `stackalloc` for small inputs is not unnecessary; it's a performance optimization. It's faster than heap allocation for small buffers, and the 512-character limit ensures we don't risk stack overflow for larger inputs.

The transition to heap allocation for larger inputs is precisely to avoid stack overflow issues while maintaining efficiency for the more common case of smaller inputs.

This approach provides a good balance between performance and safety across various input sizes.

rpallares: I just says that you Always allocate the small buffer on stack even if you don't use it.
It's cheap but still more expensive than don't allocate it AT all when you don't need it

wioniqle-q: Thank you for your insightful comment. You're absolutely right, and I appreciate your attention to optimization details. Based on your feedback, I've actually updated the implementation to address this exact issue. Here's what changed:

1. We now first check if the input length exceeds our stack allocation limit (512 characters):
 - `if (input.Length > MaxStackAllocSize) return ToTitleCaseLarge(input, culture);`
 -- If it does, we immediately use the heap-based method, avoiding any stack allocation.

2. For smaller inputs, we now allocate only the exact amount needed:
- `Span<char> initialBuffer = stackalloc char[input.Length];`

damieng:
We appreciate the discussion of performance and intent behind the PR but ToTitleCase is not actually used. It's a left-over helper method when we were considering providing a set of case conversions.

In the end only the camel case methods are used and not by default.  A user would have to manually register the CamelCaseElementNameConvention. Additionally these methods are only called once when the first DbContext is created and then the resultings strings are cached in the model metadata so any improvements to them would have negligble real-world implications.

wioniqle-q:
> We appreciate the discussion of performance and intent behind the PR but ToTitleCase is not actually used. It's a left-over helper method when we were considering providing a set of case conversions.
> 
> In the end only the camel case methods are used and not by default. A user would have to manually register the CamelCaseElementNameConvention. Additionally these methods are only called once when the first DbContext is created and then the resultings strings are cached in the model metadata so any improvements to them would have negligble real-world implications.



> We appreciate the discussion of performance and intent behind the PR but ToTitleCase is not actually used. It's a left-over helper method when we were considering providing a set of case conversions.
> 
> In the end only the camel case methods are used and not by default. A user would have to manually register the CamelCaseElementNameConvention. Additionally these methods are only called once when the first DbContext is created and then the resultings strings are cached in the model metadata so any improvements to them would have negligble real-world implications.

Thank you for the detailed feedback and clarification. I appreciate you taking the time to explain the current usage and context of these methods. So, I'm closing that PR.

src/MongoDB.EntityFrameworkCore/NamingHelperMethods.cs
rpallares: Even better 💪

wioniqle-q: > Even better 💪

Absolutely agree! But PR was not accepted, my orientation was to optimize the existing code even if it was a one-time cost. Even though DBContext caches the CamelCaseElementNameConvention when the user calls it, and even though I have hundreds of entities and the one-time creation cost is very small, my goal was to optimize directly, So it was optimizing the existing code. There are about 916 projects on Github that use the CamelCaseElementNameConvention and I thought it was commonly used. The benchmark results were already obvious. Also thanks a lot for your feedback!!

PR #155, opened 2024-10-08
Noticed a typo in the BsonTimeSpanOptionsAttributeConvention.cs

damieng:
Small silly typo crept in at some point in the past.

Technically this is a breaking change as it's public but I'd be surprised if anyone used it. Should still mention it in the 8.2 release notes/breaking changes though just to be sure.


papafe (approved):
LGTM

PR #156, opened 2024-10-08
EF-167: Fallback to using unfiltered ListDatabaseNames

damieng:
DatabaseExists/Async needs to determine if a database exists. It does this using ListDatabaseName/Async and using an appropriate name filter.

Unfortunately the name filter is not supported on some Atlas configurations so we now detect when it fails and fall back to getting the full list and filtering client side instead.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
+++ b/src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
@@ -25,7 +25,6 @@ using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Query;
 using MongoDB.Bson;
 using MongoDB.Driver;
-using MongoDB.Driver.Linq;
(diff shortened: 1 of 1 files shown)


src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
```diff
@@ -155,19 +155,48 @@ public async Task<bool> DeleteDatabaseAsync(CancellationToken cancellationToken
     /// <inheritdoc/>
     public bool DatabaseExists()
     {
-        using var cursor = _client.ListDatabaseNames(BuildListDbNameOptions());
-        return cursor.Any();
+        if (_useDatabaseNameFilter)
```
rstam, line 158: It's risky to cache information like this because if they upgrade the server without restarting the application the application is stuck forever thinking that `_useDsatabaseNameFilter == false`.

If we don't care about that then this is fine.

rstam (approved):
LGTM with a warning.

PR #159, opened 2024-10-21
EF-171: Allow type discriminators without mapped property

damieng:
Enables type discriminators to be used without a mapped property by relaxing some of the shadow property validation and ensuring type discriminator shadow properties are always serialized.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Infrastructure/MongoModelValidator.cs
+++ b/src/MongoDB.EntityFrameworkCore/Infrastructure/MongoModelValidator.cs
@@ -73,7 +73,6 @@ public class MongoModelValidator : ModelValidator
         ValidateNoUnsupportedAttributesOrAnnotations(model);
         ValidateElementNames(model);
         ValidateNoUnsupportedShadowProperties(model);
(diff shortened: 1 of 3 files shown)


src/MongoDB.EntityFrameworkCore/Infrastructure/MongoModelValidator.cs
```diff
@@ -72,7 +72,8 @@ public override void Validate(IModel model, IDiagnosticsLogger<DbLoggerCategory.
         ValidateMaximumOneRowVersionPerEntity(model);
         ValidateNoUnsupportedAttributesOrAnnotations(model);
         ValidateElementNames(model);
-        ValidateNoShadowProperties(model);
+        ValidateNoUnsupportedShadowProperties(model);
+        ValidateElementNames(model);
```
papafe, line 76: You have `ValidateElementsNames` some row earlier too, is it supposed to be like this?

damieng: No, looks like something that came out around the rebasing. I'll fix.

PR #163, opened 2024-11-21
EF-173: Support HasIndex by creating them during CreateDatabase

damieng:
## Usage

```csharp
mb.Entity<Address>().HasIndex(a => new { a.Country, a.PostCode }, "country_zip");
mb.Entity<Company>().HasIndex(c => c.OrderRef).IsUnique();
mb.Entity<Orders>().HasIndex(o => o.OrderDate).IsDescending();
mb.Entity<Documents>()
  .HasIndex(d => d.DocRef)
    .HasCreateIndexOptions(new CreateIndexOptions { Sparse = true });
```

## Features

- Single property, multi-property and string named property specification
- Explicit name and generated name indexes
- Descending and mixed ordering
- Unique indexes
- Sparse & custom options through HasCreateIndexOptions
- MQL-based partialFilter support through CreateIndexOptions<BsonDocument>

### Indexes once created are not automatically modified or deleted 

This is not a migration system. Developers will need to write manual code that uses the MongoDB C# Driver if they need to delete or modify existing indexes. 

### Unnamed indexes use generated names based on properties

As per the server the client generates a name based on the properties and sort order of those properties. This means changing either of these elements causes a new index to be created UNLESS a name has been specified. It is recommended to name indexes explicitly using `HasName`.

### Partial filters must be specified as MQL

LINQ queries can not be used to declare the partial indexes but MQL can be.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Extensions/MongoIndexBuilderExtensions.cs
+++ b/src/MongoDB.EntityFrameworkCore/Extensions/MongoIndexBuilderExtensions.cs
@@ -30,6 +30,21 @@ public static class MongoIndexBuilderExtensions
     /// </summary>
     /// <param name="indexBuilder">The builder for the index being configured.</param>
     /// <param name="options">The <see cref="CreateIndexOptions"/> for this index.</param>
+    /// <returns>A builder to further configure the index.</returns>
+    public static IndexBuilder HasCreateIndexOptions(
+        this IndexBuilder indexBuilder,
+        CreateIndexOptions? options)
+    {
+        indexBuilder.Metadata.SetCreateIndexOptions(options);
+
+        return indexBuilder;
+    }
+
+    /// <summary>
+    /// Configures the <see cref="CreateIndexOptions"/> for the index.
+    /// </summary>
+    /// <param name="indexBuilder">The builder for the index being configured.</param>
+    /// <param name="options">The <see cref="CreateIndexOptions"/> for this index.</param>
     /// <param name="fromDataAnnotation">Indicates whether the configuration was specified using a data annotation.</param>
     /// <returns>
     /// The same builder instance if the configuration was applied, <see langword="null" /> otherwise.
@@ -39,29 +54,26 @@ public static class MongoIndexBuilderExtensions
         CreateIndexOptions<BsonDocument>? options,
         bool fromDataAnnotation = false)
     {
-        if (indexBuilder.CanSetCreateIndexOptions(options, fromDataAnnotation))
+        if (!indexBuilder.CanSetCreateIndexOptions(options, fromDataAnnotation))
         {
-            indexBuilder.Metadata.SetCreateIndexOptions(options, fromDataAnnotation);
-            return indexBuilder;
+            return null;
         }
+
+        indexBuilder.Metadata.SetCreateIndexOptions(options, fromDataAnnotation);
         return indexBuilder;
     }
 
     /// <summary>
     /// Configures the <see cref="CreateIndexOptions"/> for the index.
     /// </summary>
+    /// <typeparam name="TEntity">The entity type being configured.</typeparam>
     /// <param name="indexBuilder">The builder for the index being configured.</param>
     /// <param name="options">The <see cref="CreateIndexOptions"/> for this index.</param>
-    /// <returns>
-    /// The same builder instance if the configuration was applied, <see langword="null" /> otherwise.
-    /// </returns>
+    /// <returns>A builder to further configure the index.</returns>
     public static IndexBuilder<TEntity> HasCreateIndexOptions<TEntity>(
         this IndexBuilder<TEntity> indexBuilder,
-        CreateIndexOptions? options)
-    {
-        indexBuilder.Metadata.SetCreateIndexOptions(options);
-        return indexBuilder;
-    }
+        CreateIndexOptions options)
+        => (IndexBuilder<TEntity>)HasCreateIndexOptions((IndexBuilder)indexBuilder, options);
 
     /// <summary>
     /// Returns a value indicating whether the given <see cref="CreateIndexOptions"/> can be set for the index.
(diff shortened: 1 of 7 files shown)


src/MongoDB.EntityFrameworkCore/Extensions/MongoIndexBuilderExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+public static class MongoIndexBuilderExtensions
+{
+    /// <summary>
+    /// Configures the <see cref="CreateIndexOptions"/> for the index.
+    /// </summary>
+    /// <param name="indexBuilder">The builder for the index being configured.</param>
+    /// <param name="options">The <see cref="CreateIndexOptions"/> for this index.</param>
+    /// <param name="fromDataAnnotation">Indicates whether the configuration was specified using a data annotation.</param>
+    /// <returns>
+    /// The same builder instance if the configuration was applied, <see langword="null" /> otherwise.
```
papafe, line 35: Doesn't this return the input builder nevertheless and not null? Same question for the other methods

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseCreator.cs
```diff
@@ -36,16 +37,16 @@ public bool EnsureDeleted()
...
     /// <inheritdoc/>
     public bool EnsureCreated()
-        => clientWrapper.CreateDatabase(currentDbContext.Context.Model);
+        => clientWrapper.CreateDatabase(currentDbContext.Context.GetService<IDesignTimeModel>());
 
     /// <inheritdoc/>
-    public Task<bool> EnsureCreatedAsync(CancellationToken cancellationToken = new())
-        => clientWrapper.CreateDatabaseAsync(currentDbContext.Context.Model, cancellationToken);
+    public Task<bool> EnsureCreatedAsync(CancellationToken cancellationToken = default)
+        => clientWrapper.CreateDatabaseAsync(currentDbContext.Context.GetService<IDesignTimeModel>(), cancellationToken);
```
papafe, line 49: Why this change?

src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
```diff
@@ -85,59 +85,118 @@ public IClientSessionHandle StartSession()
...
+                indexManager.CreateOne(CreateIndexDocument(index, name));
             }
         }
-
-        return !existed;
     }
 
     /// <inheritdoc />
-    public async Task<bool> CreateDatabaseAsync(IModel model, CancellationToken cancellationToken = default)
+    public async Task<bool> CreateDatabaseAsync(IDesignTimeModel model, CancellationToken cancellationToken = default)
```
papafe, line 134: Why this change?

papafe (commented):
Overall it makes sense! I just got some questions

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseCreator.cs
damieng: The design-time model contains the indexes but they are removed in the optimized run-time models.

While brekaing, the IMongoClientWrapper interface does state that this interface is for internal use and is not subject to the breaking change policy.

src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
damieng: The design-time model contains the indexes but they are removed in the optimized run-time models.

While brekaing, the IMongoClientWrapper interface does state that this interface is for internal use and is not subject to the breaking change policy.

src/MongoDB.EntityFrameworkCore/Extensions/MongoIndexBuilderExtensions.cs
damieng: It should, but this is the method signature and docs that EF's own builders use everywhere so I did not want to deviate from that.

src/MongoDB.EntityFrameworkCore/Storage/MongoDatabaseCreator.cs
papafe: Good to know!

src/MongoDB.EntityFrameworkCore/Extensions/MongoIndexBuilderExtensions.cs
papafe: In this case shouldn't we follow the conventions and return null if the configuration is not applied?

damieng: I'm just following the same patterns for extension methods used across EF - e.g. https://github.com/dotnet/efcore/blob/main/src/EFCore.Cosmos/Extensions/CosmosModelBuilderExtensions.cs#L55 which is the same sort of thing for Cosmos.

damieng: Oops, you're totally right, now fixed.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Storage/MongoClientWrapperTests.cs
```diff
@@ -43,80 +49,251 @@ public void CreateDatabase_creates_database()
...
+        var database = new TemporaryDatabaseFixture();
+        var context = MyContext.CreateCollectionOptions(database.MongoDatabase,
+            mb => mb.Entity<Address>().HasIndex("PostCode"));
+        var client = context.GetService<IMongoClientWrapper>();
+
+        _ = async
+            ? await client.CreateDatabaseAsync(context.GetService<IDesignTimeModel>())
+            : client.CreateDatabase(context.GetService<IDesignTimeModel>());
+
+        Assert.Equal(2, GetIndexes(database.MongoDatabase, "Addresses").Count);
```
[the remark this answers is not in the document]
damieng: Added.

PR #165, opened 2024-12-04
EF-178: Support Mql.Exists, IsMissing & IsMissingOrNull

damieng:
Supports using the C# Driver's `Mql.Exists`, `Mql.IsMissing` and `Mql.IsMissingOrNull` functions in EF queries with either real or shadow properties.

Notes

- IsMissing can't be used with List navigations as we do not support materializing entities with missing list navigations at this time. That will be tackled in EF-164 (Defaults) in an upcoming release.
- Current queries use $expr and $type of `missing` for now until CSHARP-5427 ships when it will becomes the more expected `$exists` operation.


damieng:
Do not review yet, is dependent on new version of C# Driver and shadow property support.

sanych-sun (approved):
LGTM

PR #169, opened 2024-12-06
DEVPROD-13322: template authenticode key name

drichmdb:
This commit templates our authenticode key name in preparation for the authenticode 2021 deprecation. The variable will be added to our Evergreen project as mongo-authenticode-2024.


sanych-sun (approved):
LGTM

PR #170, opened 2024-12-11
EF-180: Ensure operations after OfType<T> use entity serializer/EF meta

damieng:
https://jira.mongodb.org/browse/EF-180


papafe (approved):
Looks good to me for what I can understand 😄

PR #173, opened 2025-01-08
EF-182: Stop EF from defaulting key properties as ValueGenerated.OnAdd

damieng:
By default Entity Framework annotates key properties of specific types as ValueGenerated.OnAdd despite there being no EF-provided value generators to go with them by default.

In EF8 this was not a problem as people could still specify them manually but in EF9 this throws an exception if there is no generator available even though it already has a value. 

This PR changes our MongoValueGenerationConvention to no longer defer to the built-in ValueGenerationConvention for anything except Guid.

Fixes EF-182 and related to https://github.com/dotnet/efcore/issues/35422


papafe (approved):
LGTM

PR #175, opened 2025-01-09
Fix encryption tests

sanych-sun:


evergreen/evergreen.yml
```diff
@@ -60,15 +60,13 @@ functions:
 
           export MONGO_ORCHESTRATION_HOME="$DRIVERS_TOOLS/.evergreen/orchestration"
           export MONGODB_BINARIES="$DRIVERS_TOOLS/mongodb/bin"
-          export CRYPT_SHARED_LIB_PATH="$CRYPT_SHARED_LIB_PATH"
```
sanych-sun, line 63: We do not have the value yet. It will be populated after executing `bootstrap-mongo-orchestration`

evergreen/evergreen.yml
```diff
@@ -537,23 +534,14 @@ axes:
         variables:
           DRIVER_VERSION: "latest"
 
-  - id: target_runtime
```
sanych-sun, line 540: Decided to remove it for now. As we have a single target framework in the project.

evergreen/run-tests.sh
```diff
@@ -21,6 +21,4 @@ then
 fi
 
 dotnet clean "./MongoDB.EFCoreProvider.sln"
-
-
-dotnet test "./MongoDB.EFCoreProvider.sln" -e MONGODB_URI="${MONGODB_URI}" --runtime "${TARGET_RUNTIME}" --results-directory ./artifacts/test-results --logger "junit;LogFileName=TEST_{assembly}.xml;FailureBodyFormat=Verbose" --logger "console;verbosity=detailed"
```
sanych-sun, line 26: `--runtime "${TARGET_RUNTIME}"` is a wrong thing to do here, it should be `--framework` instead, but it also makes no sense as we have a single target for the project.

damieng:
LGTM as far as I understand the flow of env variables!

damieng (approved):

PR #177, opened 2025-01-20
EF-129: Add support for value converters against nullable properties

damieng:
https://jira.mongodb.org/browse/EF-129

Note: Some fluent syntax from the EF builder produces valueconverters that actually accept the nullable type even though nulls are not supposed to go through the converts (and instead traverse from one side to the other independently).

We can't use such converters so instead throw when they are encountered and suggest the alternative working syntax.


src/MongoDB.EntityFrameworkCore/Serializers/ValueConverterSerializer.cs
```diff
@@ -27,8 +29,8 @@ namespace MongoDB.EntityFrameworkCore.Serializers;
 /// <typeparam name="TStorage">The CLR type being used for storage this serializer.</typeparam>
 internal class ValueConverterSerializer<TActual, TStorage> : IBsonSerializer<TActual>
 {
-    private readonly ValueConverter<TActual, TStorage> _valueConverter;
-    private readonly IBsonSerializer<TStorage> _storageSerializer;
+    protected readonly ValueConverter<TActual, TStorage> _valueConverter;
```
papafe, line 32: Do they need to be protected now?

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ValueConverterTests.cs
```diff
@@ -257,105 +249,356 @@ class DayIsString : IdIsObjectId
...
         db.SaveChanges();
 
-        var found = database.GetCollection<DayIsString>(collection.CollectionNamespace)
-            .AsQueryable().First();
+        var found = database.GetCollection<DayIsString>(collection.CollectionNamespace).AsQueryable().First();
+        Assert.Equal(original._id, found._id);
+        Assert.Equal(day.ToString(), found.day);
+    }
+
+    private static readonly Action<ModelBuilder>? UnsupportedNullableConverter = mb =>
```
papafe, line 360: A curiosity, how come this converter is unsupported?

papafe (approved):
Just a small question, for the rest, LGTM

PR #185, opened 2025-02-13
EF-77: Support top-level scalars

damieng:
Also supports some top-level set operations like Union, Distinct, Concat and Cast.


papafe (approved):
LGTM!

PR #186, opened 2025-02-18
update to silkbomb 2.0

wratner:
Silkbomb 2.0 introduces the augment command which will allow you to upload your sbom to be scanned by OSV and have an augmented SBOM returned with any vulnerabilities in a single step.

In the case of this repo, the SBOM Lite is stored in source and is scanned and augmented nightly. This update will allow you to download the augmented SBOM. 

Let me know if I'm missing anything or if there is anything else that can be cleaned up or improved. Thanks!


adelinowona:
Tackled this here: #190

PR #193, opened 2025-03-08
EF-195: Support setting BSON representation on PropertiesConfigurationBuilder

mburumaxwell:
> Not sure whether this fits the category of allowed contributions. If there are adjustments required, I can make them.

This allows configuration of properties of a certain type to have one particular representation in the whole context as compared to having this done from one property to another.

Example:

```cs
private class Customer
{
    public int Id { get; set; }
    public DateTimeOffset CreatedAt { get; set; }
    public DateTimeOffset UpdatedAt { get; set; }
    public DateTimeOffset? ArchivedAt { get; set; }
}

private class OrdersDbContext : DbContext
{
    public DbSet<Customer> Customers { get; set; }

    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        => optionsBuilder.UseMongoDB("mongodb://localhost:27017", "mydB");

    protected override void ConfigureConventions(ModelConfigurationBuilder configurationBuilder)
    {
        base.ConfigureConventions(configurationBuilder);

        configurationBuilder.Properties<DateTimeOffset>().HaveBsonRepresentation(BsonType.DateTime);
    }
}
```

### DIFF
--- a/tests/MongoDB.EntityFrameworkCore.UnitTests/Extensions/MongoPropertiesConfigurationBuilderExtensionsTests.cs
+++ b/tests/MongoDB.EntityFrameworkCore.UnitTests/Extensions/MongoPropertiesConfigurationBuilderExtensionsTests.cs
@@ -35,6 +35,7 @@ public class MongoPropertiesConfigurationBuilderExtensionsTests
 
         var representation = property.GetBsonRepresentation();
         Assert.NotNull(representation);
+        Assert.Equal(bsonType, representation.BsonType);
     }
 
     [Theory]
--- a/tests/MongoDB.EntityFrameworkCore.UnitTests/Extensions/MongoPropertiesConfigurationBuilderExtensionsTests.cs
+++ b/tests/MongoDB.EntityFrameworkCore.UnitTests/Extensions/MongoPropertiesConfigurationBuilderExtensionsTests.cs
@@ -1,0 +1,87 @@
+﻿/* Copyright 2023-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


tests/MongoDB.EntityFrameworkCore.UnitTests/Extensions/MongoPropertiesConfigurationBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,87 @@
...
+    [InlineData(BsonType.Int64)]
+    public void HaveBsonRepresentation_can_set_type_on_property(BsonType bsonType)
+    {
+        using var db = new TestDbContext(bsonType);
+
+        var property = db.GetProperty((TestEntity t) => t.DateTimeOffsetProperty);
+        Assert.NotNull(property);
+
+        var representation = property.GetBsonRepresentation();
+        Assert.NotNull(representation);
```
damieng, line 37: Thanks for the PR, looks great, just missing the Assert.Equal(bsonType, representation.BsonType) line here.

damieng (changes requested):

tests/MongoDB.EntityFrameworkCore.UnitTests/Extensions/MongoPropertiesConfigurationBuilderExtensionsTests.cs
mburumaxwell: Thanks. Fixed it

damieng:
Fixes https://jira.mongodb.org/browse/EF-195

damieng (approved):

mburumaxwell:
@damieng it seems setting this for nullable types works in these tests but not in the actual database (e.g. `DateTimeOffset?`). Same thing happens to per property configuration. Would you happen to know where the issue is and where that would be fixed?

damieng:
Hi. It appears both HasBsonRepresenation and HaveBsonRepresentation do not apply to nullable properties. I've create a ticket in our issue tracker at https://jira.mongodb.org/browse/EF-201 and have found the root cause and potential fix.

damieng:
PR to fix at https://github.com/mongodb/mongo-efcore-provider/pull/198/files

PR #194, opened 2025-03-10
EF-196: Support configuration-level HasDateTimeKind

damieng:
Ensures the pattern in EF-195 is also available for HasDateTimeKind.


papafe (approved):
lgtm!

PR #198, opened 2025-04-08
EF-201: Apply HasBsonRepresentation to nullable properties

damieng:
Details at https://jira.mongodb.org/browse/EF-201

Breaking change for existing databases where it has not been honored.


mburumaxwell:
The sources seems it would fix it, though I cannot make much of the tests. I could test it if merged.
I believe anyone using `HasBsonRepresentation(...)` or `HaveBsonRepresentation(...)` on nullable properties would expect null or the value be in the representation. When I tested before writing the comment, I can could read the data not following the representation correctly but writing was where the issue would be. In my view, only readers with specific structure requirements such as reporting tools would suffer from this.

PR #200, opened 2025-04-12
EF-199: Allow navigation collections to be tested for nullability

damieng:
https://jira.mongodb.org/browse/EF-199

The (temporary) solution here until EF allows us to opt-out of this behavior is to wrap the property access in a cast which is sufficient enough to break the pattern matching in EF Core's NavigationExpandingExpressionVisitor.


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/OwnedEntityTests.cs
```diff
@@ -1093,4 +1170,11 @@ private record SimpleOwned
             }
         ]
     };
+
+    private static readonly FirstLevel FirstLevel1NullChildren = new()
```
papafe, line 1174: Is this used somewhere?

papafe (changes requested):
Overall it looks good, there is just one unused method I think.

papafe (approved):
LGTM

PR #206, opened 2025-06-10
EF-207: Add support for Entity Framework alternate keys

damieng:
- Adds support for [EF Alternate Keys](https://jira.mongodb.org/browse/EF-207) backed with MongoDB unique indexes
- Also fixes a [rogue exception](https://jira.mongodb.org/browse/EF-206) when EF temporarily adds multiple keys during model construciton

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Metadata/Conventions/PrimaryKeyDiscoveryConvention.cs
+++ b/src/MongoDB.EntityFrameworkCore/Metadata/Conventions/PrimaryKeyDiscoveryConvention.cs
@@ -13,7 +13,6 @@
  * limitations under the License.
  */
 
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.EntityFrameworkCore;
@@ -44,7 +43,7 @@ public class PrimaryKeyDiscoveryConvention : KeyDiscoveryConvention
     {
         var entityType = entityTypeBuilder.Metadata;
         if (entityType.BaseType != null
-            || (entityType.IsKeyless && entityType.GetIsKeylessConfigurationSource() != ConfigurationSource.Convention)
+            || entityType.IsKeyless && entityType.GetIsKeylessConfigurationSource() != ConfigurationSource.Convention
             || !entityTypeBuilder.CanSetPrimaryKey((IReadOnlyList<IConventionProperty>?)null))
         {
             return;
@@ -58,7 +57,7 @@ public class PrimaryKeyDiscoveryConvention : KeyDiscoveryConvention
             {
                 var keyProperties = ownership.Properties.ToList();
 
-                // Ones in a collection need their parent Id + an index
+                // Owned in a collection need their parent Id + an index
                 if (!ownership.IsUnique)
                 {
                     var uniqueShadowKeyProperty = entityTypeBuilder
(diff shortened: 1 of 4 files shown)


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ValueConverterTests.cs
```diff
@@ -1183,6 +1183,50 @@ class GenericPropertyEntity<T>
         public T value { get; set; }
     }
 
+    [Fact]
+    public void Converter_on_key_and_secondary_property_does_not_cause_key_errors()
```
adelinowona, line 1187: Is this test supposed to be using alternate keys?

adelinowona (changes requested):
Some comments and also is this feature meant to be primarily a "database initialization" feature? Meaning we don't plan on supporting scenarios where users try to add alternate keys for existing entities?

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Storage/MongoClientWrapperTests.cs
```diff
@@ -178,6 +178,45 @@ public async Task CreateDatabase_creates_nested_index_on_owns_many_owns_one(bool
         Assert.Equal(3, indexList.Count);
     }
 
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public async Task CreateDatabase_creates_alternate_keys(bool async)
```
adelinowona, line 184: Should we also add tests for multiple alternate keys on the same entity?

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ValueConverterTests.cs
damieng: No, this was the original bug that caused me to go check out alternate keys. It temporarily during the model building process looks like an alternate key even though it isn't - just the way EF builds it. The validation was really in the wrong place but now that validation is remove anyway with this feature and I wanted some coverage.

damieng:
We support Database.EnsureCreated() which ensures the DB and collections are created as well as any declared indexes and alternate keys by name. The user can add one later on to their code and as long as they're calling Database.EnsureCreated and there is no index with that name we'll create it even if the rest of the DB exists.

This is about the best we can do without a full migration based system which would be a huge effort beyond the EF provider and out of scope for the project.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Storage/MongoClientWrapperTests.cs
damieng: Good idea, now updated. Also made sure the names reflect any bson mapping in the test.

src/MongoDB.EntityFrameworkCore/Metadata/Conventions/PrimaryKeyDiscoveryConvention.cs
```diff
@@ -79,24 +78,18 @@ protected override void TryConfigurePrimaryKey(IConventionEntityTypeBuilder enti
...
+        // Set what we decided to use for the primary key as "_id" in the document
+        var keys = entityType.GetKeys().Where(k => k.IsPrimaryKey()).ToArray();
+        if (keys.Length == 1)
         {
-            case > 1:
-                throw new NotSupportedException("Alternate keys are not supported by the MongoDB EF Core Provider.");
-            case 1:
-                keys[0].Properties[0].SetElementName("_id");
-                break;
+            keys[0].Properties[0].SetElementName("_id");
```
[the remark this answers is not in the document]
damieng: This method runs a lot during the process. As long as we end up with one key it should be fine. We've not seen any user problems caused by this pattern. We could at some future date perhaps move it to so post-validation step.

PR #207, opened 2025-06-12
EF-208: Add ASP.NET/DI example to README.md

armanossiloko:
While attempting to set up EF Core with MongoDB in a personal project using Dependency injection, I quickly figured out that the `PlanetDbContext.Create()` example from the [README](https://github.com/mongodb/mongo-efcore-provider#getting-started) is meant only for flat out console applications or anything where people would manually instantiate the DbContext.

Since EF Core is normally used with Dependency Injection and I am really used to it, I had to take a different approach to using EF Core with MongoDB. As a result, I created a simple example showcasing what people can use instead of the default example from the Getting started section.


damieng:
HI, thanks for your PR. 

I don't think we want to check any examples or samples into the product repo as a policy. 

Did you encounter any difficulty in just following typical EF patterns for using ASP.NET in order to get the MongoDB provider working?

It's a tricky balance between pointing people in the right direction and not replicating EF documentation itself.

damieng:
You should absolutely be able to do:

```csharp
builder.Services.AddDbContext<CatalogDbContext>((provider, options) =>
{
    options.UseMongoDB(connectionString, databaseName);
});
```

We could add that to the README.md file - it's quite similar to existing ones besides the separate databaseName field.

It looks like from your example you might have needed additional settings? Or are those just for illustration.

README.md
```diff
@@ -31,15 +32,42 @@ internal class PlanetDbContext : DbContext
...
 var mongoConnectionString = Environment.GetEnvironmentVariable("MONGODB_URI");
 var mongoClient = new MongoClient(mongoConnectionString);
 var db = PlanetDbContext.Create(mongoClient.GetDatabase("planets"));
 db.Database.EnsureCreated();
+var planet = db.Planets.FirstOrDefault(x => x.Name == "Earth");
 ```
 
+#### Option 2: Using Dependency Injection
+```csharp
+var nosqlConnectionString = builder.Configuration.GetConnectionString(ConnectionStringKey)!;
```
damieng, line 48: Let's call this mongoConnectionString like the other example and maybe also not rely on ConnectionStringKey but put a name in there like MongoDbUri ?

README.md
```diff
@@ -31,15 +32,42 @@ internal class PlanetDbContext : DbContext
...
+builder.Services.AddDbContext<PlanetDbContext>(options =>
+{
+    options.UseMongoDB(nosqlConnectionString, DatabaseName);
+});
+```
+
+If you need some more configuration, you can create the `MongoClient` and inject it into the DbContext:
+
+```csharp
+var nosqlConnectionString = builder.Configuration.GetConnectionString(ConnectionStringKey)!;
```
damieng, line 58: Same here as other comment about naming.

damieng (changes requested):
Just a small tweak. I think we're going to move away recommending people creating a MongoClient themselves by default unless they have specific use-cases as it messes with disposal and the forthcoming queryable encryption but I can update that after.

Thanks!

armanossiloko:
> Just a small tweak. I think we're going to move away recommending people creating a MongoClient themselves by default unless they have specific use-cases as it messes with disposal and the forthcoming queryable encryption but I can update that after.
> 
> Thanks!

Yeah, I actually think that it'd be a better approach overall to let the framework handle the disposal of everything rather than let people create stuff on their own as most people never really dispose of anything. :D

README.md
```diff
@@ -55,7 +55,7 @@ builder.Services.AddDbContext<PlanetDbContext>(options =>
 If you need some more configuration, you can create the `MongoClient` and inject it into the DbContext:
 
 ```csharp
-var mongoConnectionString = builder.Configuration.GetConnectionString(ConnectionStringKey)!;
+var mongoConnectionString = Environment.GetEnvironmentVariable("MONGODB_URI");
```
damieng, line 58: Sorry I wasn't very clear here. I agree with you it should come from the builder.Configuration - I just think we need to put a sample name there instead of relying on a const/variable called ConnectionStringKey, e.g.

```csharp
var mongoConnectionString = builder.Configuration.GetConnectionString("MongoDbUri")!;
```

damieng (changes requested):

README.md
armanossiloko: Ah understood. Sorry, I guess I misunderstood it.

armanossiloko:
If merged, I would probably recommend squashing all commits in the process because the commit history has a lot to it.

damieng (approved):
LGMT thanks!

PR #208, opened 2025-06-18
Clean-up/refactoring of encryption tests split from EF-204

damieng:
This is the clean-up/refactoring work split out from the EF-204 Queryable Encryption work.

It does not change behavior in any way it just sets up the tests to be re-usable and does some minor re-organization of non-test code.


adelinowona (approved):
LGTM

PR #209, opened 2025-07-02
Add new UseMongo API alternatives to accepting MongoClient

damieng:
Adds new APIs to accept a MongoClientSetting instead of MongoClient as part of our work to encourange usage of MongoClientSettings or ConnectionString over providing a IMongoClient as the latter causes issues with Disposal and things like Queryable Encryption where we have to configure a MongoClient on behalf of the user.

Additionally also adds a connectionString accepting overload that does not take a databaseName argument instead drawing from the connectionStrings default database.

Fixes [EF-209](https://jira.mongodb.org/browse/EF-209)


src/MongoDB.EntityFrameworkCore/Extensions/MongoDbContextOptionsExtensions.cs
```diff
@@ -28,11 +27,11 @@ namespace Microsoft.EntityFrameworkCore;
 public static class MongoDbContextOptionsExtensions
 {
     /// <summary>
-    /// Configures the <see cref="DbContext"/> subclass to connect to a MongoDB database.
+    /// Configures the context to connect to a MongoDB database.
```
ajcvickers, line 30: Why remove the docs ref here?

tests/MongoDB.EntityFrameworkCore.UnitTests/Infrastructure/MongoOptionsExtensionTest.cs
```diff
@@ -46,18 +45,47 @@ public static void Can_set_mongo_client_and_database_name(string databaseName)
         Assert.Equal(databaseName, options.DatabaseName);
     }
 
+    [Theory]
+    [InlineData("SomeDatabase")]
+    public static void Can_set_mongo_client_settings_and_database_name(string databaseName)
```
ajcvickers, line 50: Why is this a theory when it only has one data value?

src/MongoDB.EntityFrameworkCore/Extensions/MongoDbContextOptionsExtensions.cs
damieng: Was just making them all the consistent text closer to the ones in EF.

tests/MongoDB.EntityFrameworkCore.UnitTests/Infrastructure/MongoOptionsExtensionTest.cs
damieng: I think this started as a theory with multiple values and then stopped having them. I can create a separate PR to turn them back if needed.

src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
```diff
@@ -60,7 +60,21 @@ public MongoClientWrapper(
...
+
+        var databaseName = options?.DatabaseName;
+        if (databaseName == null && options?.ConnectionString != null)
+        {
+            try
+            {
+                var connectionString = new MongoUrl(options.ConnectionString);
+                databaseName = connectionString.DatabaseName;
+            }
+            catch (FormatException)
```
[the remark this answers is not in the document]
ajcvickers: That's reasonable, but what happens if I make a typo in my connection string? When will it fail and how? (I can test this if you're not sure.)

src/MongoDB.EntityFrameworkCore/Extensions/MongoDbContextOptionsExtensions.cs
```diff
@@ -28,11 +27,11 @@ namespace Microsoft.EntityFrameworkCore;
 public static class MongoDbContextOptionsExtensions
```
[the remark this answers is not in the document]
ajcvickers: Additional context here: https://github.com/dotnet/efcore/issues/8427

src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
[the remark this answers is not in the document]
damieng: I'm not sure what the exact existing behavior is I just know I don't want to change it in this PR.

PR #211, opened 2025-07-07
EF-204: Queryable Encryption schema generation

damieng:


damieng:
The two remaining failures against latest relate to [SERVER-106469](https://jira.mongodb.org/browse/SERVER-106469) and my attempts to work around that haven't improved things.

src/MongoDB.EntityFrameworkCore/Extensions/MongoEntityTypeBuilderExtensions.cs
```diff
@@ -26,7 +26,7 @@ namespace MongoDB.EntityFrameworkCore.Extensions;
 /// </summary>
 public static class MongoEntityTypeBuilderExtensions
 {
-    const string ArgumentNameCannotBeEmptyExceptionMessage = "The string argument 'name' cannot be empty.";
+    private const string ArgumentNameCannotBeEmptyExceptionMessage = "The string argument 'name' cannot be empty.";
```
damieng, line 29: Just linter stuff.

src/MongoDB.EntityFrameworkCore/Metadata/MongoAnnotationNames.cs
```diff
@@ -55,4 +55,44 @@ public static class MongoAnnotationNames
     /// The key for create index options annotations.
     /// </summary>
     public const string CreateIndexOptions = Prefix + nameof(CreateIndexOptions);
+
+    /// <summary>
+    /// The key for the id of the data key used for encryption.
+    /// </summary>
+    public const string EncryptionDataKeyId = Prefix + nameof(EncryptionDataKeyId);
```
damieng, line 62: This one isn't prefixed with Queryable as we might add support for CSFLE which could re-use it.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerExtensions.cs
```diff
@@ -109,4 +110,105 @@ private static MqlQueryEventDefinition LogExecutedMqlQuery(IDiagnosticsLogger lo
...
+                        level,
+                        MongoEventId.RecommendedMinMaxRangeMissing,
+                        LogRecommendedMinMaxRangeMissingString)));
+        }
+
+        return (EventDefinition<string, string, string>)definition;
+    }
+
+    private const string LogRecommendedMinMaxRangeMissingString =
+        "The property '{entityType}.{propertyType}' is configured for Queryable Encryption range queries but is missing the recommended min/max values required for {bsonType} storage.";
```
ajcvickers, line 164: If there is a good docs page to go to, then adding a fwd/go/whatever short URL here pointing to it might be useful. Really just something to consider any time we have a message.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerExtensions.cs
```diff
@@ -109,4 +110,105 @@ private static MqlQueryEventDefinition LogExecutedMqlQuery(IDiagnosticsLogger lo
...
+                        level,
+                        MongoEventId.EncryptedNullablePropertyEncountered,
+                        EncryptedNullablePropertyEncounteredString)));
+        }
+
+        return (EventDefinition<string, string>)definition;
+    }
+
+    private const string EncryptedNullablePropertyEncounteredString =
+        "The property '{entityType}.{property}' is configured for Queryable Encryption but is also nullable. Null is not supported by Queryable Encryption and will throw when attempting to save.";
```
ajcvickers, line 213: I can't believe you wrote all this boilerplate code without complaining. Folks on the EF team always complained to me when they had to add a log message!

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyExtensions.cs
```diff
@@ -155,6 +145,16 @@ public static void SetBsonRepresentation(
         => property.FindAnnotation(MongoAnnotationNames.BsonRepresentation)?.GetConfigurationSource();
 
 
+    /// <summary>
+    /// Gets the <see cref="DateTimeKind"/> of the <see cref="DateTime"/> property is mapped to when targeting MongoDB.
+    /// </summary>
+    /// <param name="property">The <see cref="IReadOnlyProperty"/> to obtain the <see cref="DateTimeKind"/> for.</param>
+    public static DateTimeKind GetDateTimeKind(this IReadOnlyProperty property)
```
ajcvickers, line 152: We should chat about this. :-) (But not as part of this PR.)

src/MongoDB.EntityFrameworkCore/Infrastructure/MongoOptionsExtension.cs
```diff
@@ -96,45 +121,88 @@ public virtual MongoOptionsExtension WithDatabaseName(string databaseName)
...
     {
-        ArgumentNullException.ThrowIfNull(mongoClient);
-        EnsureConnectionNotAlreadyConfigured(nameof(MongoClient));
+        var clone = Clone();
+        clone.KmsProviders = kmsProviders;
+        return clone;
+    }
+
+    /// <summary>
+    /// Obtains the current encryption provider.
```
ajcvickers, line 159: "Gets" is normal. "Returns" is normal. "Obtains" is...weird. :-) I mean, it's fine, just unusual.

ajcvickers (approved):
Looks good. I didn't see any tests for properties with value converters defined. I think a couple would be helpful, especially given the cases we have discussed over the past week where value conversion intersects this.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerExtensions.cs
damieng: We write the docs after the feature so we could come back and update it later. Perhaps I should create a ticket to come back through the various custom exceptions we have and write a troubleshooting page that has them all on it?

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyExtensions.cs
damieng: This was existing code that was just moved in the file but yeah we can talk about that offline.

src/MongoDB.EntityFrameworkCore/Storage/MongoClientWrapper.cs
```diff
@@ -388,26 +384,95 @@ private IEnumerable<T> ExecuteScalar<T>(MongoExecutableQuery executableQuery)
...
+        var usesQueryableEncryption = queryableEncryptionSchema.Count > 0;
 
-        if (options?.MongoClientSettings != null)
-            return new MongoClient(options.MongoClientSettings);
+        var preconfiguredMongoClient = (IMongoClient?)serviceProvider.GetService(typeof(IMongoClient)) ?? options?.MongoClient;
+        if (preconfiguredMongoClient != null)
+        {
+            if (usesQueryableEncryption)
+            {
+                throw new InvalidOperationException(
```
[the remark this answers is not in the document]
damieng: If it was pre-configured and already setup with queryable encryption on the mongoclient then they wouldn't need to use any of our fluent API so that usesQueryableEncryption would be false.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerExtensions.cs
ajcvickers: The value may not be there to come back and do it later. I'll make a note to bring it up with people. (Also, if the link it a a redirection not a direct link, then the link can be created now as a placeholder and updated to point to the docs when they are done. It can even point to the general server docs for queryable encryption until then. But anyway, more just something to think about.)

PR #215, opened 2025-07-21
EF-211: Queryable Encryption server-side schema support

damieng:


src/MongoDB.EntityFrameworkCore/Extensions/MongoServiceCollectionExtensions.cs
```diff
@@ -174,7 +174,7 @@ public static IServiceCollection AddEntityFrameworkMongoDB(this IServiceCollecti
             .TryAdd<IModelRuntimeInitializer, MongoModelRuntimeInitializer>()
             .TryAddProviderSpecificServices(
                 b => b
-                    .TryAddScoped<IMongoSchemaProvider, MongoSchemaProvider>()
+                    .TryAddScoped<IQueryableEncryptionSchemaProvider, QueryableEncryptionSchemaProvider>()
```
damieng, line 177: Decided to rename this interface to be more explicit. It hasn't shipped yet.

src/MongoDB.EntityFrameworkCore/Infrastructure/MongoModelValidator.cs
```diff
@@ -261,13 +261,7 @@ private static void ValidateEntityQueryableEncryption(
             }
 
             var dataKeyId = property.GetEncryptionDataKeyId();
-            if (dataKeyId == null)
```
damieng, line 264: Allowing empty data keys if they happened to get here somehow as the shared lib throws albeit a less helpful error. Is a bit difficult to actually get here unless they set annotations manually using the direct annotation api.

src/MongoDB.EntityFrameworkCore/Infrastructure/QueryableEncryptionSchemaGenerator.cs
```diff
@@ -127,8 +127,8 @@ private static BsonDocument GenerateSchemaField(IReadOnlyProperty property, stri
         var dataKeyId = property.GetEncryptionDataKeyId();
         var fieldSchema = new BsonDocument
         {
-            { "path", prefix + property.GetElementName() },
             { "keyId", dataKeyId != null ? new BsonBinaryData(dataKeyId.Value, GuidRepresentation.Standard) : BsonNull.Value },
+            { "path", prefix + property.GetElementName() },
```
damieng, line 131: Moved path to be after keyId simply because that's the order the server returns no matter what order you send it in.

PR #217, opened 2025-07-25
EF-213: Allow default C# driver serializers for mapping of structs with simple types

ajcvickers:
Added case serializer factory to look up the serializer for simple structs in the same way that the C# driver does. Also refactored some tests as part of investigating the issue.


tests/MongoDB.EntityFrameworkCore.FunctionalTests/Mapping/ClrTypeMappingTests.cs
```diff
@@ -180,382 +180,400 @@ public ListSubclass(IEnumerable<T> items) : base(items)
...
-    {
+    [Theory]
+    [InlineData(int.MaxValue, int.MinValue)]
+    [InlineData(-1, int.MaxValue)]
+    [InlineData(0, -1)]
+    [InlineData(1, 0)]
+    [InlineData(int.MinValue, 1)]
+    public void Int32_read_update(int inserted, int updated)
+    {
+        database.DropCollection();
```
damieng, line 362: Our current pattern was that each test fixture creates it's own database and each test it's own collection if needed.
At the start of a run it clears all the previous DBs.

Rather than relying on the dropping you can just have it create a new collection passing in the args to CreateCollection.  I'd rather go this route as then you can see all the collections and databases from the last run and don't have to jump in/comment out/break-stop to get any kind of transient collection.
e.g.

```
var collection = database.CreateCollection<Int32Entity>(values: [inserted, updated]);
```

Which will create a uniquely named one that uses those args.

damieng (changes requested):
LGTM although I'd rather we pass values to createCollection for uniquely named collections than dropping/recreating.

damieng (approved):

PR #218, opened 2025-08-02
EF-27: All the Northwind query tests

ajcvickers:
This PR brings in all the remaining Northwind query tests. This brings the spec tests count up to about 4700, with about 3100 failing. Many of these failures are around various projections, giving some decent coverage for coming work.


damieng (approved):
LGTM

PR #221, opened 2025-08-11
EF-259: Support materialization of read-only properties

damieng:
Fixes [EF-259](https://jira.mongodb.org/browse/EF-259)


ajcvickers (approved):
Looks good. I have a feeling that materialization to a readonly field may require additional work, but I can't find any examples of that, so I may be remembering wrong.

PR #227, opened 2025-08-27
EF-262: Shadow key support

damieng:


src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyExtensions.cs
```diff
@@ -278,7 +284,8 @@ public static void SetQueryableEncryptionType(
         QueryableEncryptionType? queryableEncryptionType,
         bool fromDataAnnotation = false)
         => (QueryableEncryptionType?)property
-            .SetOrRemoveAnnotation(MongoAnnotationNames.QueryableEncryptionType, Check.IsDefinedOrNull(queryableEncryptionType), fromDataAnnotation)
+            .SetOrRemoveAnnotation(MongoAnnotationNames.QueryableEncryptionType, Check.IsDefinedOrNull(queryableEncryptionType),
```
ajcvickers, line 287: What's our maximum allowed line length?

src/MongoDB.EntityFrameworkCore/Extensions/MongoPropertyExtensions.cs
```diff
@@ -587,10 +593,12 @@ internal static bool IsOwnedTypeKey(this IReadOnlyProperty property)
...
-                                         && property.ClrType == typeof(int)
-                                         && (property.ValueGenerated & ValueGenerated.OnAdd) != 0
-                                         && property.GetElementName().Length == 0;
+    internal static bool IsOwnedTypeOrdinalKey(this IReadOnlyProperty property) =>
+        property.DeclaringType as IReadOnlyEntityType is { } entityType
+        && entityType.IsOwned()
+        && property.ClrType == typeof(int)
+        && (property.ValueGenerated & ValueGenerated.OnAdd) != 0
+        && property.GetElementName().Length == 0
+        && !property.IsForeignKey()
```
ajcvickers, line 602: Theoretically, an owned type PK could be configured as an FK for something else. Do we need to exclude FKs?

damieng: We should probably talk about this as I have questions...

PR #232, opened 2025-09-04
EF-265: Make provider conventions public

ajcvickers:
Also sealed DateTimeKindConvention to align with other conventions like this.


ajcvickers:
@papafe Technically it is a breaking change, but in reality inheriting from this type would have been pointless since there are no virtual members.

damieng (approved):

PR #240, opened 2025-10-02
EF-76: Flow EF metadata down through CreateGetValueExpression

ajcvickers:
That is, the IProperty or INavigation associated with the access. Types, requiredness, type-mappings, etc. are then obtained from the property.

An alias is passed in if the name in the BSON document is different from the mapped element name for the property.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Serializers/BsonSerializerFactory.cs
+++ b/src/MongoDB.EntityFrameworkCore/Serializers/BsonSerializerFactory.cs
@@ -282,7 +282,7 @@ public sealed class BsonSerializerFactory
         return CreateGenericSerializer(typeof(EnumerableInterfaceImplementerSerializer<,>), [type, itemType], childSerializer);
     }
 
-    internal static BsonSerializationInfo GetPropertySerializationInfo(IReadOnlyProperty property)
+    internal static BsonSerializationInfo GetPropertySerializationInfo(string? alias, IReadOnlyProperty property)
     {
         var serializer = CreateTypeSerializer(property);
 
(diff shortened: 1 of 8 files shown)


src/MongoDB.EntityFrameworkCore/Serializers/BsonSerializerFactory.cs
```diff
@@ -282,7 +282,7 @@ private static IBsonSerializer GetCollectionSerializer(Type type, IBsonSerialize
         return CreateGenericSerializer(typeof(EnumerableInterfaceImplementerSerializer<,>), [type, itemType], childSerializer);
     }
 
-    internal static BsonSerializationInfo GetPropertySerializationInfo(IReadOnlyProperty property)
+    internal static BsonSerializationInfo GetPropertySerializationInfo(string? alias, IReadOnlyProperty property)
```
damieng, line 285: We should make the alias optional and come after property rather than passing null in some paths.

src/MongoDB.EntityFrameworkCore/Serializers/EntitySerializer.cs
```diff
@@ -98,7 +98,7 @@ public bool TryGetMemberSerializationInfo(string memberName, out BsonSerializati
         var property = _entityType.FindProperty(memberName);
         if (property != null)
         {
-            serializationInfo = BsonSerializerFactory.GetPropertySerializationInfo(property);
+            serializationInfo = BsonSerializerFactory.GetPropertySerializationInfo(null, property);
```
damieng, line 101: Then this won't need to change. No alias, no argument.

src/MongoDB.EntityFrameworkCore/Storage/MongoUpdate.cs
```diff
@@ -203,7 +203,7 @@ internal static void WriteNonKeyProperties(IBsonWriter writer, IUpdateEntry entr
 
     private static void WriteProperty(IBsonWriter writer, object? value, IProperty property)
     {
-        var serializationInfo = BsonSerializerFactory.GetPropertySerializationInfo(property);
+        var serializationInfo = BsonSerializerFactory.GetPropertySerializationInfo(null, property);
```
damieng, line 206: Same here.

damieng (changes requested):

PR #242, opened 2025-10-07
EF-76: More work on projections involving tuple create

ajcvickers:


src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -71,17 +71,24 @@ protected override Expression VisitShapedQuery(ShapedQueryExpression shapedQuery
         // TODO: Handle select expressions and non-EF shapers more comprehensively
         if (projectedEntityType == null)
```
ajcvickers, line 72: Ultimately, we hope to remove this entirely, right? Otherwise we may end up having to keep this in sync with what is and isn't supported down the "normal" path. Do you think a TODO is okay for this?

ajcvickers:
Replaced with a more manageable series of PRs.

PR #247, opened 2025-10-08
EF-76: Visiting binary and conditional expressions

ajcvickers:
Failing spec tests skipped.


damieng (approved):
Let's just accept the two Copilot suggestions and merge this in. I'll tidy up and unify the patterns I put in here and investigate those test failures and the MQL.Field situation.

PR #249, opened 2025-10-10
EF-271: Vector index filter fields do not use element mapping

ajcvickers:


damieng (approved):
Just the typo Copilot recommends.

PR #282, opened 2026-03-04
EF-76: Select push-down support for mixed projections

damieng:
Implements [EF-76](https://jira.mongodb.org/browse/EF-76)

## Problem

Select projections that mix entity references with scalar properties (e.g., `Select(p => new { p.Name, Entity = p })`) would previously fail. The query compiler had two rigid paths: 

1. Pure entity (returns `BsonDocument`, shaped client-side using EF shapers)
2. Pure scalar (fully pushed down to LINQ V3 and materialized outside EF)

Any `.Select` attempting to project both the entity and a scalar threw an exception.

## Solution

This PR introduces a introduces a **`ProjectionAnalyzer`** that walks the shaper expression tree to classify projections as fully push-downable or mixed. For mixed projections, a new **`MongoMixedProjectionBindingRemovingExpressionVisitor`** strips the `Select` so the server returns full `BsonDocument`s, then reads scalars from the root document using serialization info while delegating entity materialization to the existing base visitor.

It also extends `MongoProjectionBindingExpressionVisitor` to recognize `EF.Property()` and `Mql.Field()` as scalar property accesses in projection mappings and refactors `MongoShapedQueryCompilingExpressionVisitor` to share translation logic via a common `TranslateQuery<TSource>` helper across all three paths (entity, projected, mixed).

## Testing

- ~1000 lines of new functional tests covering anonymous types, DTOs, tuples, conditional expressions, `EF.Property`, and entity + scalar combinations
- Multiple EF specification tests unblocked (previously expected to throw)
- Some remaining EF specification tests still fail as they have moved on to the next issue which is mostly about subqueries and cross-collection selects which are to be handled in ticket [EF-117](https://jira.mongodb.org/browse/EF-117)

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoMixedProjectionBindingRemovingExpressionVisitor.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoMixedProjectionBindingRemovingExpressionVisitor.cs
@@ -1,0 +1,138 @@
+/* Copyright 2023-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using System.Linq.Expressions;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Query;
+using MongoDB.Bson;
+using MongoDB.EntityFrameworkCore.Query.Expressions;
+using MongoDB.EntityFrameworkCore.Storage;
+
+namespace MongoDB.EntityFrameworkCore.Query.Visitors;
+
+/// <summary>
+/// Extends <see cref="MongoProjectionBindingRemovingExpressionVisitor"/> to handle mixed projections
+/// (containing both entity references and scalar properties). In this path, the LINQ V3 query
+/// returns full BsonDocuments (Select is stripped), and scalars are read from the root document
+/// using the property's actual serialization info rather than the projection alias.
+/// </summary>
+internal sealed class MongoMixedProjectionBindingRemovingExpressionVisitor
+    : MongoProjectionBindingRemovingExpressionVisitor
+{
+    private readonly MongoQueryExpression _queryExpression;
+    private readonly IEntityType _rootEntityType;
+    private readonly ParameterExpression _docParameter;
+
+    public MongoMixedProjectionBindingRemovingExpressionVisitor(
+        IEntityType rootEntityType,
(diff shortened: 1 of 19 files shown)


src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoMixedProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            // Handle EF.Property<T>(entity, "propertyName") method calls
+            if (methodCall.Method.IsEFPropertyMethod()
+                && methodCall.Arguments[1] is ConstantExpression { Value: string propertyName })
+            {
+                var property = _rootEntityType.FindProperty(propertyName);
+                return (property, property != null ? null : propertyName);
+            }
+
+            // Handle Mql.Field<TDoc, TField>(entity, "fieldName", serializer) calls
```
ajcvickers, line 127: What happens if the application specifies a specific serializer? I assume we just pass it through to the driver?

src/MongoDB.EntityFrameworkCore/Query/Visitors/ProjectionAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+                return ContainsEntityReference(lambdaExpression.Body);
+
+            case ProjectionBindingExpression:
+            case ConstantExpression:
+            case ParameterExpression:
+            case DefaultExpression:
+                return false;
+
+            default:
+                // Unknown expression types conservatively prevent push-down to avoid
```
ajcvickers, line 109: Should we do something in debug builds to indicate we hit an unknown expression?

tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/VectorSearchMongoTestBase.cs
```diff
@@ -482,12 +482,41 @@ public virtual async Task VectorSearch_with_projection_of_entity_and_score(bool
...
-                _ = async ? await queryable.ToListAsync() : queryable.ToList()))
-            .Message);
+        var results = async ? await queryable.ToListAsync() : queryable.ToList();
+        Assert.Equal(2, results.Count);
+
+        Assert.Equal("Entity Framework Core in Action", results[0].Book.Title);
+        Assert.Equal("Jon P Smith", results[0].Book.Author);
+        Assert.True(results[0].Score > 0);
+
+        Assert.Equal("Programming Entity Framework: DbContext", results[1].Book.Title);
```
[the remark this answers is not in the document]
ajcvickers: With this data, the ordering should always be the same. If it isn't, it indicates a bug.

ajcvickers (approved):

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoMixedProjectionBindingRemovingExpressionVisitor.cs
damieng: I don't know what we do about that in any Mql.Field before or after this PR. I've raised EF-309 to track investigating (and possibly correcting) the behavior and then documenting it.

src/MongoDB.EntityFrameworkCore/Query/Visitors/ProjectionAnalyzer.cs
damieng: Added a Debug.Assert

ajcvickers (approved):

PR #293, opened 2026-03-20
EF-202: Entity comparisons

damieng:
## Summary

Adds support for direct entity-to-entity comparisons in LINQ queries by automatically rewriting them into property-level comparisons. Previously, any direct entity comparison (e.g. `Where(p => p == someEntity)`) would throw `NotSupportedException` for regular entities or silently (and incorrectly) fail through the MongoDB driver's serializer for owned entities which performed incorrect comparisons given the lack of provided EF metadata.

### Entity comparison strategy

**Non-owned entities** are compared by their primary key properties. A query like `Where(p => p == entity)` is rewritten to `Where(p => p._id == entity._id)`. Composite keys are fully supported with all key properties AND'd together.

**Owned entities** are compared by all CLR-mapped properties (excluding shadow properties like the FK back to the owner). For example, comparing a `Location` owned entity with `latitude` and `longitude` rewrites to `latitude == loc.latitude && longitude == loc.longitude`.

### Supported patterns

| Pattern | Example |
|---|---|
| `==` / `!=` in Where/First | `Where(p => p == entity)`, `First(p => p.location != loc)` |
| `Equals` method forms | `entity.Equals(other)`, `object.Equals(a, b)` |
| `Contains` on queryable | `dbSet.Contains(entity)` |
| `Contains` on owned collection | `p.locations.Contains(location)` |
| `!Contains` (negated) | `!p.locations.Contains(location)` |
| Local array contains | `Where(p => localArray.Contains(p))` |
| `Any` with entity equality | `p.locations.Any(l => l == location)` |
| Null comparisons | `Where(p => p == null)` |
| List with nulls | `localList.Contains(p)` where list includes null entries |

### Implementation

All entity comparison rewriting lives in `MongoEFToLinqTranslatingExpressionVisitor`:

- `VisitBinary` intercepts `==` / `!=` operators
- `VisitContainsMethod` intercepts `Contains` calls (instance, `Enumerable.Contains`, `Queryable.Contains`)
- `Equals` method calls are intercepted in the existing `Visit` switch
- Shared helpers (`FindEntityType`, `GetComparisonProperties`, `CreateEntityEqualityExpression`, `CombineBalanced`) eliminate duplication across all rewrite paths
- Expression trees are combined into balanced binary trees to avoid deep left-leaning nesting


src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -398,16 +415,292 @@ void ThrowForBadOptions(string reason)
...
+
+            var isQueryable = collection.Type.IsGenericType
+                              && collection.Type.GetGenericTypeDefinition() == typeof(IQueryable<>);
+            var anySource = isQueryable ? typeof(Queryable) : typeof(Enumerable);
+            var anyMethod = anySource.GetMethods()
+                .First(m => m.Name == nameof(Enumerable.Any) && m.GetParameters().Length == 2)
+                .MakeGenericMethod(elementType);
+
+            return Expression.Call(null, anyMethod, collection,
+                isQueryable ? Expression.Quote(lambda) : lambda);
```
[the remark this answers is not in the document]
damieng: yagni

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -398,16 +415,295 @@ void ThrowForBadOptions(string reason)
...
+        }
+
+        return base.VisitBinary(node);
+    }
+
+    /// <summary>
+    /// Find an entity type in the model, preferring owned types (which require searching all entity types).
+    /// </summary>
+    private IEntityType? FindEntityType(Type clrType)
+        => _queryContext.Context.Model.GetEntityTypes().FirstOrDefault(e => e.ClrType == clrType && e.IsOwned())
```
ajcvickers, line 485: This may not return the correct entity type. Consider case where non-owned entity has CustomerAddress and HomeAddress properties, both of type "Address". These will be different owned types in the model, but this code will arbitrarily choose one of them.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -398,16 +415,295 @@ void ThrowForBadOptions(string reason)
...
+    /// <summary>
+    /// Get the properties to use for entity comparison: primary key properties for non-owned entities,
+    /// all CLR-mapped properties for owned entities.
+    /// Returns null if no usable properties are found.
+    /// </summary>
+    private static IReadOnlyList<IProperty>? GetComparisonProperties(IEntityType entityType)
+    {
+        if (entityType.IsOwned())
+        {
+            var properties = entityType.GetProperties()
```
ajcvickers, line 497: Should this always exclude shadow properties? I agree that they should be excluded when comparing to an instance, but when both sides of the comparison are translated, then it seems like shadow properties should be included. I'm not sure if this method is used in those cases or not.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -398,16 +415,295 @@ void ThrowForBadOptions(string reason)
...
+        bool isOwned, ExpressionType nodeType = ExpressionType.Equal)
+    {
+        var comparisons = properties.Select(p =>
+        {
+            Expression queryAccess;
+            Expression constantExpr;
+
+            if (isOwned)
+            {
+                queryAccess = p.PropertyInfo != null
```
ajcvickers, line 521: I don't think this respects the configuration for field access. We should read from the field when configured as such, regardless of whether the property exists or not.

ajcvickers: Thinking about this more, I don't think this is possible with the current LINQ implementation because it would require the driver to respect this too. Also, it only matters when actually reading values from instances.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -398,16 +415,295 @@ void ThrowForBadOptions(string reason)
...
+
+        return CombineBalanced(comparisons, nodeType == ExpressionType.Equal
+            ? Expression.AndAlso
+            : Expression.OrElse);
+    }
+
+    /// <summary>
+    /// Combine a list of expressions into a balanced binary tree using the given combiner.
+    /// </summary>
+    private static Expression CombineBalanced(IReadOnlyList<Expression> expressions,
```
ajcvickers, line 547: 👍

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/CompositeKeyQueryTests.cs
```diff
@@ -56,6 +56,43 @@ public void Should_query_by_key_component_with_gt()
         Assert.All(result, e => Assert.True(e.Key2 > 2));
     }
 
+    [Fact]
+    public void Should_match_entity_equal()
```
ajcvickers, line 60: Do we support any queries where both sides of the comparison are translated?

ajcvickers (changes requested):

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
damieng: Discussed, will add a new model validator to ensure consistency between configurations.

damieng: Now included

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/CompositeKeyQueryTests.cs
damieng: We do not nor do we support constant to constant.

ajcvickers (approved):

PR #295, opened 2026-04-10
Update test-all skill to use parallel sub-agents with model selection

damieng:


.claude/skills/test-all/SKILL.md
```diff
@@ -1,134 +1,58 @@
 ---
 name: test-all
```
sanych-sun, line 2: I do not want to be a party killer, but shouldn't we have this a little other way: have a bash script to run the tests (sure Claude can write that for us), output tests results into a specific file. If test run is failed - call Claude to investigate, summarize, etc.

damieng: If we run it via a bash script then it does it sequentially unless you want to get into trying to ruin them in parallel and then combining the results. This way does it in parallel using lightweight models/agents that can anayze the failures.

sanych-sun: Sure we can run a command in background and wait for exit code. And if there are any failures we can call Claude with necessary model and ask to analyze the test result(s).

damieng: I don't think you can "call claude" after running a script. Claude has to be in control from the start.

ajcvickers (approved):

PR #296, opened 2026-04-16
EF-312: Logging, cast and SingleOrDefault fixes identified by Claude

damieng:
Fixes [EF-312](https://jira.mongodb.org/browse/EF-312)


src/MongoDB.EntityFrameworkCore/Storage/ValueConversion/Decimal128ToDecimalConverter.cs
```diff
@@ -54,5 +54,5 @@ public Decimal128ToDecimalConverter(ConverterMappingHints? mappingHints = null)
     /// A <see cref="ValueConverterInfo" /> for the default use of this converter.
     /// </summary>
     public static ValueConverterInfo DefaultInfo { get; }
-        = new(typeof(string), typeof(ObjectId), i => new Decimal128ToDecimalConverter(i.MappingHints));
```
ajcvickers, line 57: I think this means that value converter selection is not being tested, or not fully. I've made a note of this to check later, and will file a Jira ticket if needed. It may also be covered by spec tests when we get them in.

ajcvickers (approved):

PR #298, opened 2026-05-01
EF-313: Rewrite binding for robustness with non-trivial shapes

damieng:
## Problem

Two distinct bugs, fixed here together because they share the same code path.

### Bug 1 — aliased scalar projections use the wrong serializer

A projection like `new { Alias = e.Prop }` pushes a `$project` stage to MongoDB that emits the field under the alias key. When the shaper reads it back, it looked up `Alias` as a *property name* on the root entity type in order to find a serializer. Three failure modes depending on what `Alias` happens to collide with:

| Alias collides with… | Outcome |
|---|---|
| no EF property at all | `InvalidOperationException: The property '<Root>.Alias' could not be found` |
| a different EF property with incompatible nullability | `ArgumentException` — Expression.Call rejects a `Nullable<T>` where `T` is expected |
| a different EF property with compatible type | query runs but uses the **wrong** serializer — silent data corruption |

The same bug affects properties on owned entities and owned collections.

### Bug 2 — `NoTrackingWithIdentityResolution` collapses distinct owned entities

When reading owned entities in `NoTracking` mode, the shaper needs a synthetic placeholder key (the owned type has no `_id` of its own — its identity comes from its principal's key). Previously, this synthetic key was emitted for *all* non-tracking modes, including `NoTrackingWithIdentityResolution`.

`NoTrackingWithIdentityResolution` uses a standalone state manager that deduplicates entity instances by key. With every owned entity sharing the same synthetic key, two owners with distinct owned entities would collapse into one — the second owner's entity silently returned the first's data.

### Aliased serializer resolution

`VisitExtension` now walks the projection's source expression (via `TryResolveFieldAccess`, promoted from the mixed visitor to the base class) to find the originating `IProperty`. It then reads the BSON element by alias key but deserializes using that property's own serializer and nullability metadata.

A new `BsonBinding.CreateGetValueExpression(doc, elementKey, sourceProperty, type)` overload separates the lookup key from the property metadata so the two concerns can no longer collide.

The mixed-projection path (a `Select` that contains both entity references and scalar properties, which LINQ V3 cannot push down) is also corrected: the pushed-down `Select` is now stripped from the LINQ expression before execution so the driver returns full entity `BsonDocument`s; the client-side shaper then handles the projection.

### Synthetic key scoping

`_useSyntheticOwnedKeys` is now set only when `trackingBehavior == QueryTrackingBehavior.NoTracking`. `NoTrackingWithIdentityResolution` queries receive real principal keys so the standalone state manager can deduplicate correctly.

## Commit structure

The PR is split into four commits to make review easier:

1. **Refactor** — widen `bool trackQueryResults` to `QueryTrackingBehavior` enum throughout the visitor chain (no behavior change; prerequisite for commit 4's scoping logic)
2. **Refactor** — rename `ProjectionBindings` → `_projectionBindings` and `_ordinalParameterBindings` → `_ordinalMappings` (no behavior change; reduces noise in commit 3's diff)
3. **Fix** — core EF-313 projection fix: `TryResolveFieldAccess`, new `BsonBinding` overload, mixed-path `StripPushedDownSelect`, tests for aliased owned and root properties
4. **Fix** — restrict synthetic owned-entity keys to `NoTracking` only; regression test covering `NoTrackingWithIdentityResolution`

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
@@ -43,9 +43,9 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
     private readonly ParameterExpression DocParameter;
     private readonly bool _trackQueryResults;
     private readonly Dictionary<ParameterExpression, Expression> _materializationContextBindings = new();
-    private readonly Dictionary<Expression, ParameterExpression> ProjectionBindings = new();
+    private readonly Dictionary<Expression, ParameterExpression> _projectionBindings = new();
     private readonly Dictionary<Expression, (IEntityType EntityType, Expression BsonDocExpression)> _ownerMappings = new();
-    private readonly Dictionary<Expression, Expression> _ordinalParameterBindings = new();
+    private readonly Dictionary<Expression, Expression> _ordinalMappings = new();
     private List<IncludeExpression> _pendingIncludes = [];
 
     /// <summary>
@@ -98,15 +98,15 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
                             throw new InvalidOperationException(CoreStrings.TranslationFailed(extensionExpression.Print()));
                     }
 
-                    var bsonArray = ProjectionBindings[objectArrayProjection];
+                    var bsonArray = _projectionBindings[objectArrayProjection];
                     var jObjectParameter = Expression.Parameter(typeof(BsonDocument), bsonArray.Name + "Object");
                     var ordinalParameter = Expression.Parameter(typeof(int), bsonArray.Name + "Ordinal");
 
                     var accessExpression = objectArrayProjection.InnerProjection.ParentAccessExpression;
-                    ProjectionBindings[accessExpression] = jObjectParameter;
+                    _projectionBindings[accessExpression] = jObjectParameter;
                     _ownerMappings[accessExpression] =
                         (objectArrayProjection.Navigation.DeclaringEntityType, objectArrayProjection.AccessExpression);
-                    _ordinalParameterBindings[accessExpression] = Expression.Add(
+                    _ordinalMappings[accessExpression] = Expression.Add(
                         ordinalParameter, Expression.Constant(1, typeof(int)));
 
                     var innerShaper = (BlockExpression)Visit(collectionShaperExpression.InnerShaper);
@@ -200,14 +200,14 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
                     if (projectionExpression is ObjectArrayProjectionExpression objectArrayProjectionExpression)
                     {
                         innerAccessExpression = objectArrayProjectionExpression.AccessExpression;
-                        ProjectionBindings[objectArrayProjectionExpression] = parameterExpression;
+                        _projectionBindings[objectArrayProjectionExpression] = parameterExpression;
                         fieldName ??= objectArrayProjectionExpression.Name;
                     }
                     else
                     {
                         var entityProjectionExpression = (EntityProjectionExpression)projectionExpression;
                         var accessExpression = entityProjectionExpression.ParentAccessExpression;
-                        ProjectionBindings[accessExpression] = parameterExpression;
+                        _projectionBindings[accessExpression] = parameterExpression;
                         fieldName ??= entityProjectionExpression.Name;
 
                         switch (accessExpression)
@@ -340,7 +340,7 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
                 var ownership = entityType.FindOwnership();
                 if (ownership?.IsUnique == false && property.IsOwnedTypeOrdinalKey())
                 {
-                    var readExpression = _ordinalParameterBindings[docExpression];
+                    var readExpression = _ordinalMappings[docExpression];
                     if (readExpression.Type != type)
                     {
                         readExpression = Expression.Convert(readExpression, type);
@@ -417,7 +417,7 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
         };
 
         var innerExpression = docExpression;
-        if (ProjectionBindings.TryGetValue(docExpression, out var innerVariable))
+        if (_projectionBindings.TryGetValue(docExpression, out var innerVariable))
         {
             innerExpression = innerVariable;
         }
--- a/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
@@ -54,20 +54,17 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
     /// <param name="rootEntityType">The <see cref="IEntityType"/> this projection relates to.</param>
     /// <param name="queryExpression">The <see cref="MongoQueryExpression"/> this visitor should use.</param>
     /// <param name="docParameter">The parameter that will hold the <see cref="BsonDocument"/> input parameter to the shaper.</param>
-    /// <param name="trackQueryResults">
-    /// <see langword="true"/> if the results from this query are being tracked for changes,
-    /// <see langword="false"/> if they are not.
-    /// </param>
+    /// <param name="trackingBehavior">The <see cref="QueryTrackingBehavior"/> for this query.</param>
     public MongoProjectionBindingRemovingExpressionVisitor(
         IEntityType rootEntityType,
         MongoQueryExpression queryExpression,
         ParameterExpression docParameter,
-        bool trackQueryResults)
+        QueryTrackingBehavior trackingBehavior)
     {
         _queryExpression = queryExpression;
         _rootEntityType = rootEntityType;
         DocParameter = docParameter;
-        _trackQueryResults = trackQueryResults;
+        _trackQueryResults = trackingBehavior == QueryTrackingBehavior.TrackAll;
     }
 
     protected override Expression VisitExtension(Expression extensionExpression)
--- a/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
@@ -26,6 +26,7 @@ using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Query;
 using Microsoft.EntityFrameworkCore.Storage;
 using MongoDB.Bson;
+using MongoDB.Driver;
 using MongoDB.EntityFrameworkCore.Extensions;
 using MongoDB.EntityFrameworkCore.Query.Expressions;
 using MongoDB.EntityFrameworkCore.Storage;
@@ -75,6 +76,39 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
                 {
                     var projection = GetProjection(projectionBindingExpression);
 
+                    // Alias is null when the projection's expression has no natural name —
+                    // the BsonDoc itself is the value.
+                    if (projection.Alias is null)
+                    {
+                        return DocParameter;
+                    }
+
+                    // Resolve the source IProperty so we apply its serializer / nullability —
+                    // not whatever EF property happens to share the alias name on the root entity.
+                    var fieldAccess = TryResolveFieldAccess(projection.Expression);
+                    if (fieldAccess.Property != null)
+                    {
+                        return BsonBinding.CreateGetValueExpression(
+                            DocParameter,
+                            projection.Alias,
+                            fieldAccess.Property,
+                            projectionBindingExpression.Type);
+                    }
+
+                    // For genuinely non-property expressions (arithmetic, constants, Mql.Field) the
+                    // push-down result document has the alias as its BSON element name — read it raw.
+                    // When projection.Expression is null (EF key-property binding via BindProperty)
+                    // there is no source expression to inspect; fall through to the property-lookup
+                    // path so the alias is resolved against the declared entity type, which gives us
+                    // the correct BSON element name (e.g. OrderID → _id).
+                    if (projection.Expression != null)
+                    {
+                        return BsonBinding.CreateGetElementValue(
+                            DocParameter,
+                            projection.Alias,
+                            projectionBindingExpression.Type);
+                    }
+
                     return CreateGetValueExpression(
                         DocParameter,
                         projection.Alias,
@@ -104,11 +138,9 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
 
                     var accessExpression = objectArrayProjection.InnerProjection.ParentAccessExpression;
                     _projectionBindings[accessExpression] = jObjectParameter;
-                    _ownerMappings[accessExpression] =
-                        (objectArrayProjection.Navigation.DeclaringEntityType, objectArrayProjection.AccessExpression);
-                    _ordinalMappings[accessExpression] = Expression.Add(
-                        ordinalParameter, Expression.Constant(1, typeof(int)));
-
+                    _ownerMappings[accessExpression] = (objectArrayProjection.Navigation.DeclaringEntityType, objectArrayProjection.AccessExpression);
+                    _ownerMappings[jObjectParameter] = (objectArrayProjection.Navigation.DeclaringEntityType, objectArrayProjection.AccessExpression);
+                    _ordinalMappings[accessExpression] = Expression.Add(ordinalParameter, Expression.Constant(1, typeof(int)));
                     var innerShaper = (BlockExpression)Visit(collectionShaperExpression.InnerShaper);
 
                     innerShaper = AddIncludes(innerShaper);
@@ -201,7 +233,7 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
                     {
                         innerAccessExpression = objectArrayProjectionExpression.AccessExpression;
                         _projectionBindings[objectArrayProjectionExpression] = parameterExpression;
-                        fieldName ??= objectArrayProjectionExpression.Name;
+                        fieldName ??= FindProjectionAlias(objectArrayProjectionExpression) ?? objectArrayProjectionExpression.Name;
                     }
                     else
                     {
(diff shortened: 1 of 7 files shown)


ajcvickers:
@damieng This is the output from the provider reviewer skill I am working on. I will go through this along with my manual review and make specific comments where I think Claude is right. But posting here so you can see the results:

# PR review: #298 — EF-313: Rewrite binding for robustness with non-trivial shapes

URL: https://github.com/mongodb/mongo-efcore-provider/pull/298
diff range: upstream/main...90cb03de6e6db377dab335be0d294758d1385a15

## PR summary

**Description**: This PR touches the Query area — specifically the projection-binding-removal visitors (`MongoProjectionBindingRemovingExpressionVisitor`, `MongoMixedProjectionBindingRemovingExpressionVisitor`, `MongoShapedQueryCompilingExpressionVisitor`) and the `BsonBinding` storage helper that produces deserialization expressions. Behaviorally, it fixes two distinct shaper bugs that share the same code path: (1) aliased scalar projections like `Select(e => new { Alias = e.Prop })` were looking the alias up as a property *name* on the root entity, so depending on what the alias collided with users got a "property not found" exception, a nullability `ArgumentException`, or — worst — a silent serializer mismatch on a same-typed unrelated property; the fix walks the source expression to recover the originating `IProperty` and feeds its serializer/nullability into a new `BsonBinding.CreateGetValueExpression` overload that separates the lookup key from the property metadata. (2) The synthetic owned-entity key the shaper emits for no-tracking reads was applied to `NoTrackingWithIdentityResolution` as well, and that mode's standalone state manager deduplicated all owned entities into one — so two owners' distinct owned values collapsed; the fix scopes the synthetic key to plain `NoTracking`, enabled by widening an internal `bool trackQueryResults` to the `QueryTrackingBehavior` enum throughout the visitor chain. The change is a focused bug-fix (~574/-223 over 7 files), split into two prep refactors plus two fixes for reviewer convenience, with new functional tests (aliased scalar/owned/owned-collection projections, `NoTrackingWithIdentityResolution` owned-entity regression) and the bonus that nine previously-broken spec tests in `NorthwindMiscellaneousQueryMongoTest` now pass or throw cleaner exceptions. No EF-version conditionals are touched, so multi-EF coverage is plausible.

**Assessment**: Both problems are clearly real — the alias-collision case is a silent-data-corruption hazard and the `NoTrackingWithIdentityResolution` case produces wrong results, so both warrant fixing. Pairing them is justified because they live in the same shaper path and one's refactor (the enum widening) enables the other's scoping logic. The approach — resolving the source `IProperty` via the projection expression rather than re-looking up by alias name — is the structurally right move and matches what the underlying mistake was. Scope feels right: refactors are small and pre-isolated into their own commits, and adjacent code (mixed-projection path stripping the pushed-down `Select`) is fixed where the same root cause applies. Tests are appropriate for a bug fix — regression tests for each failure mode at root, owned, and owned-collection levels, plus the spec-suite cleanups give incidental coverage — though I didn't see an explicit MQL-assertion test on the new alias path, which the query-reviewer may flag. No `BREAKING-CHANGES.md` entry, which is defensible here since the prior behavior was broken (exception or wrong data), not a documented contract; an api-stability reviewer may still want a note. No multi-EF `#if` red flags spotted.

**Verdict**: looks good

## Summary

7 reviewers ran (3 area+feature + 3 cross-cutting + 1 PR summary). 3 approved, 3 flagged, 0 escalated. (PR-summary verdict not counted.)

## Escalations

None.

## Cross-cutting findings

### api-stability-reviewer — flag

- `BREAKING-CHANGES.md:1` — [external-action][substantive] PR introduces a user-observable behavior change for `NoTracking` owned-entity queries (synthetic placeholder values like `ObjectId("000000000000000000000001")`, `"__mongo_ef_synthetic_owner_key"`, `1`, `1L` are now returned in place of the real owner `_id` when reading the owned entity's principal key/FK), and `BREAKING-CHANGES.md` has no entry. — Ask the author to add a section describing the new synthetic-key contract for `NoTracking` so users who read shadow FKs / owner `_id` via owned-entity projections aren't surprised.

AV: I don't think this is needed; this is a bug fix.

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs:805` — [external-action][substantive] `CreateSyntheticKeyValue` silently falls through to `Expression.Default(type)` for any principal-key CLR type outside `ObjectId/Guid/string/int/long` (e.g., `decimal`, `short`, `Guid?` boxed differently, custom value-converted types) — leaving zero/null as the "synthetic" key risks identity collisions identical to the bug 90cb03d fixed; either expand the supported set or throw a clear `NotSupportedException` so it surfaces in a test rather than at runtime.

AV: I agree with this.

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs:419` — [external-action][nit] The synthetic-key trigger compares `principalProperty.Name == "_id"` by element name string — if a user configures a non-default principal-key element name (e.g., owner with `HasElementName("ownerId")` on the PK), the synthetic-key short-circuit silently won't apply; confirm this is intentional or key off `IsPrimaryKey()` instead so behavior doesn't depend on element-naming convention.

AV: In EF terms, this is exactly right. It's possible we don't yet support non-primary principal keys, but at some point we should, so I think it's worth making this correct now.

(No public-API surface changes: `BsonBinding` stays `internal`, all three visitor classes stay `internal`, no `MongoAnnotationNames` or `MongoEventId` touched, `QueryTrackingBehavior` widening is on `internal` ctor signatures only.)

### ef-conformance-reviewer — approve

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs:70` — [external-action][substantive] The `_useSyntheticOwnedKeys` flag is set only for `NoTracking`; `TrackAll` and `NoTrackingWithIdentityResolution` both rely on reading the principal's key from `ownerBsonDocExpression` for tracked-owned-entity FK fixup, which is a change in code path coverage that warrants a multi-EF run — invoke `/test-all` to confirm EF8/EF9/EF10 all materialize owned entities correctly under all three `QueryTrackingBehavior` values.

AV: Not sure I agree it is needed, but /test-all is good anyway.

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs:418` — [fix-in-code][nit] `CreateSyntheticKeyValue` gates on `principalProperty.Name == "_id"` (string literal), but EF Core's `IProperty.Name` is the CLR member name — a user-defined PK named `Id` or `OrderId` mapped to `_id` via `HasElementName` would bypass the synthetic-key path entirely on EF9/EF10 where convention defaults could shift — replace with a check on the principal property's role (e.g. `principalProperty.IsPrimaryKey()` plus root-entity-document check) rather than a CLR-name string.

AV: I agree with this.

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs:128` — [fix-in-code][nit] The `Func<ParameterExpression, QueryTrackingBehavior, ExpressionVisitor>` signature now carries an EF Core enum across an internal delegate; this compiles cleanly under EF8/EF9/EF10 (the enum is stable), but for symmetry with the existing `#if EF8 || EF9` block at line 130 a comment noting the enum's cross-version stability would help future-proof the file — not blocking.

AV: Not very useful.

### security-reviewer — clean

(No findings. Secret-pattern grep over the diff returned zero hits; no logging, exception, or `ToString()` paths were introduced that surface BSON content, MQL, parameters, connection strings, or KMS material; the new `BsonBinding.CreateGetValueExpression(IProperty, ...)` overload and `GetPropertyValueAtElement<T>` route through model-derived element names and serializer metadata. No `MongoOptionsExtension`, `SanitizeConnectionStringForLogging`, `LogFragment`, `ShouldLogSensitiveData`, `MongoCredential`, `SslSettings`, `KmsProviders`, or `Diagnostics/` files are touched.)

## Area + feature findings

### query-reviewer — flag

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs:415` — [external-action][substantive] `_useSyntheticOwnedKeys` short-circuit for `NoTracking` keys on the synthetic-zero path triggers only when `principalProperty.Name == "_id"` (the CLR property name), so users with conventional `Id` / `OrderID` keys mapped to `_id` via `GetElementName()` silently fall through to reading the real key while users whose CLR property is literally named `_id` (the test fixtures' style) take the synthetic path — same query, different code path. — Compare against the BSON element name (`principalProperty.GetElementName() == "_id"`) or, better, drop the heuristic entirely and key the synthetic decision on `principalProperty.IsPrimaryKey()` of the owner.

AV: I agree.

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs:174` — [external-action][substantive] `StripPushedDownSelect` only un-wraps the 1-arg `Single/First/Last(Or-Default)` overloads above the pushed-down `Select`, so a captured chain ending in `Single(predicate)` / `First(predicate)` / `Count` / `Any` over a `Select` would leave the `Select` in place and mis-shape the driver-LINQ query. — Add coverage for or explicitly reject these terminator overloads, and add an MQL-asserting test in `tests/.../FunctionalTests/Query/` for `Select(...).First(predicate)` against an owned-entity to lock the contract.

AV: I agree that we should add test coverage for these and then either make sure it works or add explicit blocking exceptions.

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs:577` — [external-action][substantive] `TryResolveFieldAccessSource`'s `ParameterExpression` fallback does a `Model.GetEntityTypes().FirstOrDefault(e => e.ClrType == p.Type)` scan, returning the first match — for models with inheritance / shared-type entities this can resolve to the wrong entity and thus the wrong serializer. — Either restrict the lookup to the root entity type / its derived types, consult `_ownerMappings` first, or assert uniqueness.

AV: This may not be valid because we prevent owned types from having different mappings, but we should check.

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs:411` — [external-action][substantive] `CreateSyntheticKeyValue` covers only `ObjectId/Guid/string/int/long` and silently falls back to `Expression.Default(type)` for others (e.g. `decimal`, `short`, `byte[]`, custom value-converted PKs); under `NoTracking` the value is never observed by the StateManager, but the silent default makes any future caller who reads this synthetic value (e.g. a tracked include retrofit) see a colliding zero. — At minimum throw, log, or fall back to the real-key read for unsupported types.

AV: Same issue as raised by the stability reviewer.

- `src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs:578` — [external-action][nit] Per area `AGENTS.md`: the per-call `Mql.Field` reflection lookup is fine today (only one overload exists in driver 3.x), but a future driver overload would make `GetMethod(name, BindingFlags)` throw `AmbiguousMatchException` at class load. — Consider the explicit-signature `GetMethod(name, types)` form to future-proof against driver additions.

AV: We should fix this; I've been bitten so many times by this kind of thing in the past!

### storage-reviewer — clean

- `src/MongoDB.EntityFrameworkCore/Storage/BsonBinding.cs:215` — [fix-in-code][nit] `GetPropertyValueAtElement<T>` clones `BsonSerializationInfo` only to swap the element name; the existing serializer's `ElementPath` (if any) is dropped silently — currently safe because `GetPropertySerializationInfo` returns `ElementPath == null` for stored properties, but the rewrite is more robust if it asserts `ElementPath is null` or copies the path. — Add `Debug.Assert(serializationInfo.ElementPath is null)` to make the contract explicit.

AV: Seems okay to me.

- `src/MongoDB.EntityFrameworkCore/Storage/BsonBinding.cs:88` — [fix-in-code][nit] The new public `CreateGetValueExpression(...,IProperty,Type)` and the original five-arg overload duplicate the `BsonArray` / `BsonDocument` short-circuit logic, increasing drift risk. — Extract the array/document branches into a shared private helper or have the new overload delegate to the existing one with `property.DeclaringType` / `!property.IsNullable`.

AV: I think this is a good idea.

### spec-conformance-reviewer — flag

- `tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/NorthwindMiscellaneousQueryMongoTest.cs:2491-2497` — [fix-in-code][substantive] Eight `Select_expression_*_add_*` overrides now assert `Assert.ThrowsAsync<Xunit.Sdk.EqualException>` with no `// Fails:` rationale and no `AssertMql(...)`, breaking the established spec-conformance pattern (compare `NorthwindKeylessEntitiesQueryMongoTest`: comment + `ThrowsAsync<EqualException>` + `AssertMql`). The override silently confirms upstream equality fails but no longer documents *why* it produces a wrong result or what MQL was generated. — Add a `// Fails: <reason, e.g. driver $dateAdd unit mismatch>` comment and an `AssertMql(...)` capturing the produced pipeline for each of the eight affected tests (lines ~2491, 2494, 2497, 2500, 2503, 2506-7, 2509-10, 4189).

AV: Not sure if this is needed or not.

- `tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/NorthwindMiscellaneousQueryMongoTest.cs:2473` — [fix-in-code][substantive] `ToString_with_formatter_is_evaluated_on_the_client` now matches `Assert.Contains("Expression not supported: o.OrderID.ToString(")` — pinning a driver-side exception message including a parameter name (`o.OrderID`) makes the assertion brittle to driver upgrades and to LINQ-rewrite changes. — Match on a shorter stable substring (`"Expression not supported"`) and keep an `// Fails:` rationale comment.

AV: Agree.

- `tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/NorthwindMiscellaneousQueryMongoTest.cs:2475` — [fix-in-code][nit] Uses fully-qualified `MongoDB.Driver.Linq.ExpressionNotSupportedException`; the `MongoDB.Driver.Linq` namespace is already imported on line 22. — Drop the namespace prefix to match house style.

AV: Seems legit, but needs verification.

- `tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/NorthwindMiscellaneousQueryMongoTest.cs:2491` — [fix-in-code][nit] New `Xunit.Sdk.EqualException` references are fully qualified despite `using Xunit.Sdk;` already being present at line 25 (other files use bare `EqualException`). — Use bare `EqualException`.

AV: Seems legit, but needs verification.

- `tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/NorthwindMiscellaneousQueryMongoTest.cs:2571` — [external-action][substantive] `Select_expression_date_add_milliseconds_large_number_divided` now asserts `FormatException` "Cannot deserialize a 'Int32' from BsonType 'DateTime'" with no `// Fails:` annotation; the failure mode looks like an unsafe translation that returns a DateTime where Int32 was projected. — Confirm with the Query reviewer whether this should be `AssertTranslationFailed` instead, and add a rationale comment + JIRA reference.

AV: Need to look in more detail.

## Unmapped changes

None.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -372,6 +411,14 @@ protected Expression CreateGetValueExpression(
 
                     if (ownerBsonDocExpression != null)
                     {
+                        if (_useSyntheticOwnedKeys
```
ajcvickers: This check seems suspicious. Are we able to assume that the principal property name is "_id", or even that the the principal property is mapped to an element named "_id", since it might not be the PK? I'm not sure what the reasoning is here.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -372,6 +411,14 @@ protected Expression CreateGetValueExpression(
...
+                            && property.DeclaringType is IEntityType { } declaringEntityType
+                            && declaringEntityType.IsOwned()
+                            && principalProperty.Name == "_id")
+                        {
+                            return CreateSyntheticKeyValue(principalProperty, type);
+                        }
+
                         return CreateGetValueExpression(ownerBsonDocExpression, principalProperty, type);
```
ajcvickers, line 412: What are the valid and invalid cases for hitting this condition? Do we have tests that cover this? It seems like this might only be valid for non-generated key values, but I'm not 100% sure.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -439,6 +486,104 @@ protected Expression CreateGetValueExpression(
...
+        if (expression != null && _ownerMappings.TryGetValue(expression, out var ownerInfo))
+        {
+            return (ownerInfo.EntityType, ownerInfo.BsonDocExpression);
+        }
+
+        return (null, null);
+    }
+
+    private static readonly MethodInfo MqlFieldMethodInfo =
+        typeof(Mql).GetMethod(nameof(Mql.Field), BindingFlags.Public | BindingFlags.Static)!;
```
ajcvickers, line 569: If `Mql.Field` ever gets a new overload, then people using this code against the new driver will start seeing exceptions. This hit us several times over the years in places where we said, "Oh, there will never be another overload there..."

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -660,6 +805,42 @@ private static readonly MethodInfo PopulateCollectionMethodInfo
...
+            return value.Type == type ? value : Expression.Convert(value, type);
+        }
+
+        if (nonNullableType == typeof(long))
+        {
+            var value = Expression.Constant(1L);
+            return value.Type == type ? value : Expression.Convert(value, type);
+        }
+
+        return Expression.Default(type);
```
ajcvickers: Shouldn't this throw for types we don't support?

tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/NorthwindMiscellaneousQueryMongoTest.cs
```diff
@@ -2445,139 +2445,51 @@ public override async Task Query_expression_with_to_string_and_contains(bool asy
...
-    {
-        // Fails: Client eval in final projection EF-250
-        Assert.Contains(
-            "The property 'Order.ShipName' could not be found.",
-            (await Assert.ThrowsAsync<InvalidOperationException>(() => base.Select_expression_other_to_string(async))).Message);
-
-        AssertMql(
-        );
-    }
+        => await base.Select_expression_other_to_string(async);
```
ajcvickers, line 2462: All these tests should assert MQL.

tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/NorthwindMiscellaneousQueryMongoTest.cs
```diff
@@ -2445,139 +2445,51 @@ public override async Task Query_expression_with_to_string_and_contains(bool asy
...
-    }
+        => await base.Select_expression_int_to_string(async);
 
     public override async Task ToString_with_formatter_is_evaluated_on_the_client(bool async)
     {
-        // Fails: Client eval in final projection EF-250
         Assert.Contains(
-            "The property 'Order.ShipName' could not be found.",
-            (await Assert.ThrowsAsync<InvalidOperationException>(() =>
+            "Expression not supported: o.OrderID.ToString(",
```
ajcvickers, line 2456: Minor, but in general, consider only asserting on "Expression not supported" since this is more likely to be stable over time and we probably don't want to react to just the form of the expression not supported changing.

ajcvickers (changes requested):

PR #302, opened 2026-05-26
EF-313: Rewrite binding for robustness with non-trivial shapes

damieng:
## Problem

A projection like `new { Alias = e.Prop }` pushes a `$project` stage to MongoDB that emits the field under the alias key. When the shaper reads it back, it looked up `Alias` as a *property name* on the root entity type in order to find a serializer. Three failure modes depending on what `Alias` happens to collide with:

| Alias collides with… | Outcome |
|---|---|
| no EF property at all | `InvalidOperationException: The property '<Root>.Alias' could not be found` |
| a different EF property with incompatible nullability | `ArgumentException` — Expression.Call rejects a `Nullable<T>` where `T` is expected |
| a different EF property with compatible type | query runs but uses the **wrong** serializer — silent data corruption |

The same bug affects properties on owned entities and owned collections.

## Fix

`VisitExtension` now walks the projection's source expression (via `TryResolveFieldAccess`, promoted from the mixed visitor to the base class) to find the originating `IProperty`. It then reads the BSON element by alias key but deserializes using that property's own serializer and nullability metadata.

A new `BsonBinding.CreateGetValueExpression(doc, elementKey, sourceProperty, type)` overload separates the lookup key from the property metadata so the two concerns can no longer collide.

The mixed-projection path (a `Select` that contains both entity references and scalar properties, which LINQ V3 cannot push down) is also corrected: the pushed-down `Select` is now stripped from the LINQ expression before execution so the driver returns full entity `BsonDocument`s; the client-side shaper then handles the projection.

## Commit structure

The PR is split into four commits to make review easier:

1. **Refactor** — widen `bool trackQueryResults` to `QueryTrackingBehavior` enum throughout the visitor chain (no behavior change)
2. **Refactor** — rename `ProjectionBindings` → `_projectionBindings` and `_ordinalParameterBindings` → `_ordinalMappings` (no behavior change; reduces noise in commit 3's diff)
3. **Fix** — core EF-313 projection fix: `TryResolveFieldAccess`, new `BsonBinding` overload, mixed-path `StripPushedDownSelect`, tests for aliased owned and root properties

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
@@ -43,9 +43,9 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
     private readonly ParameterExpression DocParameter;
     private readonly bool _trackQueryResults;
     private readonly Dictionary<ParameterExpression, Expression> _materializationContextBindings = new();
-    private readonly Dictionary<Expression, ParameterExpression> ProjectionBindings = new();
+    private readonly Dictionary<Expression, ParameterExpression> _projectionBindings = new();
     private readonly Dictionary<Expression, (IEntityType EntityType, Expression BsonDocExpression)> _ownerMappings = new();
-    private readonly Dictionary<Expression, Expression> _ordinalParameterBindings = new();
+    private readonly Dictionary<Expression, Expression> _ordinalMappings = new();
     private List<IncludeExpression> _pendingIncludes = [];
 
     /// <summary>
@@ -98,15 +98,15 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
                             throw new InvalidOperationException(CoreStrings.TranslationFailed(extensionExpression.Print()));
                     }
 
-                    var bsonArray = ProjectionBindings[objectArrayProjection];
+                    var bsonArray = _projectionBindings[objectArrayProjection];
                     var jObjectParameter = Expression.Parameter(typeof(BsonDocument), bsonArray.Name + "Object");
                     var ordinalParameter = Expression.Parameter(typeof(int), bsonArray.Name + "Ordinal");
 
                     var accessExpression = objectArrayProjection.InnerProjection.ParentAccessExpression;
-                    ProjectionBindings[accessExpression] = jObjectParameter;
+                    _projectionBindings[accessExpression] = jObjectParameter;
                     _ownerMappings[accessExpression] =
                         (objectArrayProjection.Navigation.DeclaringEntityType, objectArrayProjection.AccessExpression);
-                    _ordinalParameterBindings[accessExpression] = Expression.Add(
+                    _ordinalMappings[accessExpression] = Expression.Add(
                         ordinalParameter, Expression.Constant(1, typeof(int)));
 
                     var innerShaper = (BlockExpression)Visit(collectionShaperExpression.InnerShaper);
@@ -200,14 +200,14 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
                     if (projectionExpression is ObjectArrayProjectionExpression objectArrayProjectionExpression)
                     {
                         innerAccessExpression = objectArrayProjectionExpression.AccessExpression;
-                        ProjectionBindings[objectArrayProjectionExpression] = parameterExpression;
+                        _projectionBindings[objectArrayProjectionExpression] = parameterExpression;
                         fieldName ??= objectArrayProjectionExpression.Name;
                     }
                     else
                     {
                         var entityProjectionExpression = (EntityProjectionExpression)projectionExpression;
                         var accessExpression = entityProjectionExpression.ParentAccessExpression;
-                        ProjectionBindings[accessExpression] = parameterExpression;
+                        _projectionBindings[accessExpression] = parameterExpression;
                         fieldName ??= entityProjectionExpression.Name;
 
                         switch (accessExpression)
@@ -340,7 +340,7 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
                 var ownership = entityType.FindOwnership();
                 if (ownership?.IsUnique == false && property.IsOwnedTypeOrdinalKey())
                 {
-                    var readExpression = _ordinalParameterBindings[docExpression];
+                    var readExpression = _ordinalMappings[docExpression];
                     if (readExpression.Type != type)
                     {
                         readExpression = Expression.Convert(readExpression, type);
@@ -417,7 +417,7 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
         };
 
         var innerExpression = docExpression;
-        if (ProjectionBindings.TryGetValue(docExpression, out var innerVariable))
+        if (_projectionBindings.TryGetValue(docExpression, out var innerVariable))
         {
             innerExpression = innerVariable;
         }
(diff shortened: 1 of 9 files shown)


src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -78,6 +76,39 @@ protected override Expression VisitExtension(Expression extensionExpression)
...
+                    // Resolve the source IProperty so we apply its serializer / nullability —
+                    // not whatever EF property happens to share the alias name on the root entity.
+                    var fieldAccess = TryResolveFieldAccess(projection.Expression);
+                    if (fieldAccess.Property != null)
+                    {
+                        return BsonBinding.CreateGetValueExpression(
+                            DocParameter,
+                            projection.Alias,
+                            fieldAccess.Property,
+                            projectionBindingExpression.Type);
```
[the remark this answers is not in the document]
ajcvickers: Agree with Copilot here.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -54,20 +55,17 @@ internal class MongoProjectionBindingRemovingExpressionVisitor : ExpressionVisit
...
         MongoQueryExpression queryExpression,
         ParameterExpression docParameter,
-        bool trackQueryResults)
+        QueryTrackingBehavior trackingBehavior)
     {
         _queryExpression = queryExpression;
         _rootEntityType = rootEntityType;
         DocParameter = docParameter;
-        _trackQueryResults = trackQueryResults;
+        _trackQueryResults = trackingBehavior == QueryTrackingBehavior.TrackAll;
```
ajcvickers, line 68: Are there tests for this path for NoTrackingWithIdentityResolution?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -78,6 +76,52 @@ protected override Expression VisitExtension(Expression extensionExpression)
...
+                            projectionBindingExpression.Type);
+                    }
+
+                    // For genuinely non-property expressions (arithmetic, constants, Mql.Field) the
+                    // push-down result document has the alias as its BSON element name — read it raw.
+                    // When projection.Expression is null (EF key-property binding via BindProperty)
+                    // there is no source expression to inspect; fall through to the property-lookup
+                    // path so the alias is resolved against the declared entity type, which gives us
+                    // the correct BSON element name (e.g. OrderID → _id).
+                    if (projection.Expression != null)
```
ajcvickers: Add a test where projection.Expression is null. Or if it can't be null, then adjust this code.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -439,6 +489,109 @@ protected Expression CreateGetValueExpression(
...
+
+        if (expression != null && _ownerMappings.TryGetValue(expression, out var ownerInfo))
+        {
+            return (ownerInfo.EntityType, ownerInfo.BsonDocExpression);
+        }
+
+        return (null, null);
+    }
+
+    private static readonly MethodInfo MqlFieldMethodInfo =
```
ajcvickers, line 584: Harden this against new overloads being added.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
```diff
@@ -78,6 +76,52 @@ protected override Expression VisitExtension(Expression extensionExpression)
...
+                    // outer type could differ from the property's CLR type. In practice it never
+                    // does at this call site: aliased projections that introduce a Convert (e.g.
+                    // `new { X = (long)p.intProp }`) go through the LINQ V3 push-down path and
+                    // never reach this visitor — see ProjectionAnalyzer.CanPushDown. The assert
+                    // below pins the invariant so a future change that violates it fails loudly
+                    // rather than mis-deserialising via the wrong property's serializer.
+                    var fieldAccess = TryResolveFieldAccess(projection.Expression);
+                    if (fieldAccess.Property != null)
+                    {
+                        System.Diagnostics.Debug.Assert(
```
ajcvickers: Should we not throw here?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -157,6 +156,35 @@ private MethodCallExpression CompileShapedQuery(
...
+        {
+            return captured;
+        }
+
+        if (call.Method.Name == nameof(Queryable.Select) && call.Arguments.Count == 2)
+        {
+            return call.Arguments[0];
+        }
+
+        if (call.Method.IsGenericMethod
```
ajcvickers, line 171: This seems like a fairly limited list of patterns to match. What about things like array access?

tests/MongoDB.EntityFrameworkCore.SpecificationTests/Query/NorthwindMiscellaneousQueryMongoTest.cs
```diff
@@ -2445,139 +2445,75 @@ public override async Task Query_expression_with_to_string_and_contains(bool asy
...
-    {
-        // Fails: Client eval in final projection EF-250
-        Assert.Contains(
-            "The property 'Order.ShipName' could not be found.",
-            (await Assert.ThrowsAsync<InvalidOperationException>(() => base.Select_expression_long_to_string(async))).Message);
-
-        AssertMql(
-        );
-    }
+        => await base.Select_expression_long_to_string(async);
```
ajcvickers: Baseline the MQL.

ajcvickers (changes requested):

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
damieng: Not yet, this was just changing the bool into an enum for the later piece of work that isn't on this branch. This refactoring should be a no-op.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
damieng: I've not been able to figure out a test that would get array access here and Claude is convinced it's not possible. Do you have a query that would trigger it?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
damieng: Not been able to figure out a test that would return null here but if we don't check it the compiler gets upset as technically the API can return it.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
damieng: Is now `.Single(m => m.Name == nameof(Mql.Field) && m.GetParameters().Length == 3);`

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
damieng: Done.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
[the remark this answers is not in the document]
damieng: Done.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingRemovingExpressionVisitor.cs
ajcvickers: But if it's null, then won't var fieldAccess = TryResolveFieldAccess(projection.Expression); always throw before we even get here? I would expect the null path to be flagged as heuristically unreachable by Rider. What am I missing?

damieng: Yeah, looks like it was dead code. Removed.

ajcvickers (approved):

PR #306, opened 2026-06-01
EF-117: Include

ajcvickers:
Testing PR on GitHub


BREAKING-CHANGES.md
```diff
@@ -6,6 +6,28 @@ In order to evolve the provider as we introduce new features, we will be using t
 
 ## Breaking changes in 8.4.0 / 9.1.0 / 10.0.0
 
+### Unsupported cross-collection `Include` shapes now throw `NotSupportedException`
```
ajcvickers, line 9: I don't think we need this. The change in exception type is not significant in this case.

PR #309, opened 2026-06-05
EF-117: Cross-collection Include / navigations / joins

ajcvickers:
EF-117: Cross-collection Include / navigations / joins

- Cross-collection Include/ThenInclude/navigation support built on Damien Guard's POC (his commits preserved).                                                                                                    
- Reference Include → driver-native LeftJoin; collection Include + multi-hop → $lookup/$unwind.                  
-  ⚠️ Breaking: a navigated type with a DbSet is now a separate collection                                                                                         
- Cross-collection query translation is EF10-only                                                                                                        
- A filtered Include or a query filter on a cross-collection target now fails loudly instead silent data loss     
- Spec and functional suites are green on EF8, EF9, and EF10.                                                                                                                                                                                                                                                    
- `MongoRelationshipDiscoveryConvention.ShouldBeOwnedType(...)` behavior change.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoQueryExpression.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoQueryExpression.cs
@@ -29,6 +29,8 @@ internal sealed class MongoQueryExpression : Expression
 {
     private Dictionary<ProjectionMember, Expression> _projectionMapping = new();
     private readonly List<ProjectionExpression> _projection = [];
+    private readonly List<LookupExpression> _pendingLookups = [];
+    private readonly Dictionary<IEntityType, MongoCollectionExpression> _innerCollections = new();
 
     /// <summary>
     /// Create a <see cref="MongoQueryExpression"/> for the given entity type.
@@ -86,6 +88,67 @@ internal sealed class MongoQueryExpression : Expression
 
     public IReadOnlyList<ProjectionExpression> Projection
         => _projection;
+
+    /// <summary>
+    /// Pending $lookup stages for cross-collection collection Include operations.
+    /// </summary>
+    public IReadOnlyList<LookupExpression> PendingLookups
+        => _pendingLookups;
+
+    /// <summary>
+    /// Register a $lookup stage for a cross-collection collection Include.
+    /// </summary>
+    public void AddLookup(LookupExpression lookup)
+    {
+        if (!_pendingLookups.Any(l => l.As == lookup.As))
+        {
+            _pendingLookups.Add(lookup);
+        }
+    }
+
+    /// <summary>
+    /// Inner collections involved in join operations.
+    /// </summary>
+    public IReadOnlyDictionary<IEntityType, MongoCollectionExpression> InnerCollections
+        => _innerCollections;
+
+    /// <summary>
+    /// Whether this query involves join operations across multiple collections.
+    /// </summary>
+    public bool IsJoinQuery => _innerCollections.Count > 0;
+
+    /// <summary>
+    /// Whether this join query uses the driver's _outer/_inner field naming convention.
+    /// True for Include-generated LeftJoins (LeftJoinResult), false for explicit user Joins.
+    /// </summary>
+    public bool UsesDriverJoinFields { get; set; }
+
+
+    /// <summary>
+    /// Register an inner collection for a join operation.
+    /// </summary>
+    /// <param name="entityType">The <see cref="IEntityType"/> of the inner collection.</param>
+    /// <returns>The <see cref="MongoCollectionExpression"/> for the inner collection.</returns>
+    public MongoCollectionExpression AddInnerCollection(IEntityType entityType)
+    {
+        if (!_innerCollections.TryGetValue(entityType, out var collection))
+        {
+            collection = new MongoCollectionExpression(entityType);
+            _innerCollections[entityType] = collection;
+        }
+
+        return collection;
+    }
+
+    /// <summary>
+    /// Replace a projection expression at the given index.
+    /// Used to rebind entity projections for cross-collection $lookup results.
+    /// </summary>
+    public void ReplaceProjectionAt(int index, Expression newExpression)
+    {
+        var existing = _projection[index];
+        _projection[index] = new ProjectionExpression(newExpression, existing.Alias, existing.Required);
+    }
 
     public void ApplyProjection()
     {
--- a/src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoQueryExpression.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoQueryExpression.cs
@@ -25,12 +25,10 @@ namespace MongoDB.EntityFrameworkCore.Query.Expressions;
 /// <summary>
 /// Represents a top-level MongoDB-specific collection for querying server-side.
 /// </summary>
-internal sealed class MongoQueryExpression : Expression
+internal sealed partial class MongoQueryExpression : Expression
 {
     private Dictionary<ProjectionMember, Expression> _projectionMapping = new();
     private readonly List<ProjectionExpression> _projection = [];
-    private readonly List<LookupExpression> _pendingLookups = [];
-    private readonly Dictionary<IEntityType, MongoCollectionExpression> _innerCollections = new();
 
     /// <summary>
     /// Create a <see cref="MongoQueryExpression"/> for the given entity type.
@@ -88,67 +86,6 @@ internal sealed class MongoQueryExpression : Expression
 
     public IReadOnlyList<ProjectionExpression> Projection
         => _projection;
-
-    /// <summary>
-    /// Pending $lookup stages for cross-collection collection Include operations.
-    /// </summary>
-    public IReadOnlyList<LookupExpression> PendingLookups
-        => _pendingLookups;
-
-    /// <summary>
-    /// Register a $lookup stage for a cross-collection collection Include.
-    /// </summary>
-    public void AddLookup(LookupExpression lookup)
-    {
-        if (!_pendingLookups.Any(l => l.As == lookup.As))
-        {
-            _pendingLookups.Add(lookup);
-        }
-    }
-
-    /// <summary>
-    /// Inner collections involved in join operations.
-    /// </summary>
-    public IReadOnlyDictionary<IEntityType, MongoCollectionExpression> InnerCollections
-        => _innerCollections;
-
-    /// <summary>
-    /// Whether this query involves join operations across multiple collections.
-    /// </summary>
-    public bool IsJoinQuery => _innerCollections.Count > 0;
-
-    /// <summary>
-    /// Whether this join query uses the driver's _outer/_inner field naming convention.
-    /// True for Include-generated LeftJoins (LeftJoinResult), false for explicit user Joins.
-    /// </summary>
-    public bool UsesDriverJoinFields { get; set; }
-
-
-    /// <summary>
-    /// Register an inner collection for a join operation.
-    /// </summary>
-    /// <param name="entityType">The <see cref="IEntityType"/> of the inner collection.</param>
-    /// <returns>The <see cref="MongoCollectionExpression"/> for the inner collection.</returns>
-    public MongoCollectionExpression AddInnerCollection(IEntityType entityType)
-    {
-        if (!_innerCollections.TryGetValue(entityType, out var collection))
-        {
-            collection = new MongoCollectionExpression(entityType);
-            _innerCollections[entityType] = collection;
-        }
-
-        return collection;
-    }
-
-    /// <summary>
-    /// Replace a projection expression at the given index.
-    /// Used to rebind entity projections for cross-collection $lookup results.
-    /// </summary>
-    public void ReplaceProjectionAt(int index, Expression newExpression)
-    {
-        var existing = _projection[index];
-        _projection[index] = new ProjectionExpression(newExpression, existing.Alias, existing.Required);
-    }
 
     public void ApplyProjection()
     {
--- a/src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 // Derived from EFCore.Cosmos ObjectAccessExpression.
@@ -7,85 +7,58 @@ using System;
 using System.Linq.Expressions;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Query;
-using MongoDB.EntityFrameworkCore.Extensions;
 
 namespace MongoDB.EntityFrameworkCore.Query.Expressions;
 
 /// <summary>
-/// Represents access to an object within the BsonDocument result tree.
+/// Represents access to an object within the BsonDocument result tree. The concrete subtype depends on
+/// whether the access is described by an <see cref="INavigation"/>
+/// (<see cref="NavigationObjectAccessExpression"/>) or by an <see cref="IEntityType"/> with no navigation,
+/// as for an explicit cross-collection Join (<see cref="EntityTypeObjectAccessExpression"/>).
 /// </summary>
-internal sealed class ObjectAccessExpression : Expression, IPrintableExpression, IAccessExpression
+internal abstract class ObjectAccessExpression : Expression, IPrintableExpression, IAccessExpression
 {
     /// <summary>
-    /// Create a <see cref="ObjectAccessExpression"/> for an embedded entity navigation.
+    /// Create an <see cref="ObjectAccessExpression"/>.
     /// </summary>
-    /// <param name="navigation">The <see cref="INavigation"/> this object access relates to.</param>
     /// <param name="accessExpression">The <see cref="Expression"/> of the parent containing the object.</param>
     /// <param name="required">
     /// <see langword="true"/> if this object is required,
     /// <see langword="false"/> if it is optional.
     /// </param>
-    /// <exception cref="InvalidOperationException"></exception>
-    public ObjectAccessExpression(
-        INavigation navigation,
-        Expression accessExpression,
-        bool required)
-        : this(navigation, accessExpression, required,
-            navigation.TargetEntityType.GetContainingElementName() ??
-            throw new InvalidOperationException(
-                $"Navigation '{navigation.DeclaringEntityType.DisplayName()}.{navigation.Name}' doesn't point to an embedded entity."))
-    {
-    }
-
-    /// <summary>
-    /// Create a <see cref="ObjectAccessExpression"/> with an explicit field name.
-    /// Used for cross-collection $lookup results.
-    /// </summary>
-    /// <param name="navigation">The <see cref="INavigation"/> this object access relates to.</param>
-    /// <param name="accessExpression">The <see cref="Expression"/> of the parent containing the object.</param>
-    /// <param name="required">Whether this object is required.</param>
-    /// <param name="name">The explicit field name to access in the document.</param>
-    public ObjectAccessExpression(
-        INavigation navigation,
+    /// <param name="name">The field name to access in the document.</param>
+    protected ObjectAccessExpression(
         Expression accessExpression,
         bool required,
         string name)
     {
-        Name = name;
-        Navigation = navigation;
         AccessExpression = accessExpression;
         Required = required;
+        Name = name;
     }
 
     /// <summary>
-    /// Create a <see cref="ObjectAccessExpression"/> for a cross-collection join result
-    /// where no navigation is defined (explicit Join).
+    /// The <see cref="INavigation"/> this object access relates to, or <see langword="null"/> when the
+    /// access is not described by a navigation.
     /// </summary>
-    public ObjectAccessExpression(
-        IEntityType entityType,
-        Expression accessExpression,
-        bool required,
-        string name)
-    {
-        Name = name;
-        EntityType = entityType;
-        AccessExpression = accessExpression;
-        Required = required;
-    }
+    public virtual INavigation? Navigation
+        => null;
 
-    internal IEntityType? EntityType { get; }
+    /// <summary>
+    /// The <see cref="IEntityType"/> this object access relates to when there is no navigation, otherwise
+    /// <see langword="null"/>.
+    /// </summary>
+    internal virtual IEntityType? EntityType
+        => null;
 
     /// <inheritdoc />
     public override ExpressionType NodeType
         => ExpressionType.Extension;
 
     /// <inheritdoc />
-    public override Type Type
-        => Navigation?.ClrType ?? EntityType!.ClrType;
+    public abstract override Type Type { get; }
 
     public string Name { get; }
-
-    public INavigation? Navigation { get; }
 
     public Expression AccessExpression { get; }
 
@@ -95,12 +68,12 @@ internal sealed class ObjectAccessExpression : Expression, IPrintableExpression,
     protected override Expression VisitChildren(ExpressionVisitor visitor)
         => Update(visitor.Visit(AccessExpression));
 
-    public ObjectAccessExpression Update(Expression outerExpression)
-        => outerExpression != AccessExpression
-            ? (Navigation != null
-                ? new ObjectAccessExpression(Navigation, outerExpression, Required)
-                : new ObjectAccessExpression(EntityType!, outerExpression, Required, Name))
-            : this;
+    /// <summary>
+    /// Create a copy of this expression with the given parent access expression, or return this
+    /// instance unchanged when the parent access expression is unchanged.
+    /// </summary>
+    /// <param name="outerExpression">The <see cref="Expression"/> of the parent containing the object.</param>
+    public abstract ObjectAccessExpression Update(Expression outerExpression);
 
     void IPrintableExpression.Print(ExpressionPrinter expressionPrinter)
         => expressionPrinter.Append(ToString());
@@ -113,15 +86,21 @@ internal sealed class ObjectAccessExpression : Expression, IPrintableExpression,
     public override bool Equals(object? obj)
         => obj != null
            && (ReferenceEquals(this, obj)
-               || (obj is ObjectAccessExpression objectAccessExpression
+               || (obj.GetType() == GetType()
+                   && obj is ObjectAccessExpression objectAccessExpression
                    && Equals(objectAccessExpression)));
 
-    private bool Equals(ObjectAccessExpression objectAccessExpression)
-        => Navigation == objectAccessExpression.Navigation
+    /// <summary>
+    /// Compare the members common to all <see cref="ObjectAccessExpression"/> subtypes. The caller has
+    /// already established that <paramref name="objectAccessExpression"/> is of the same runtime type, so
+    /// overrides can safely cast to add their distinguishing member.
+    /// </summary>
+    protected virtual bool Equals(ObjectAccessExpression objectAccessExpression)
+        => Name == objectAccessExpression.Name
            && AccessExpression.Equals(objectAccessExpression.AccessExpression)
            && Required == objectAccessExpression.Required;
 
     /// <inheritdoc />
     public override int GetHashCode()
-        => HashCode.Combine(Navigation, AccessExpression, Required);
+        => HashCode.Combine(Name, AccessExpression, Required);
 }
--- a/src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
@@ -17,7 +17,7 @@ namespace MongoDB.EntityFrameworkCore.Query.Expressions;
 internal sealed class ObjectAccessExpression : Expression, IPrintableExpression, IAccessExpression
 {
     /// <summary>
-    /// Create a <see cref="ObjectAccessExpression"/>.
+    /// Create a <see cref="ObjectAccessExpression"/> for an embedded entity navigation.
     /// </summary>
     /// <param name="navigation">The <see cref="INavigation"/> this object access relates to.</param>
     /// <param name="accessExpression">The <see cref="Expression"/> of the parent containing the object.</param>
@@ -30,15 +30,50 @@ internal sealed class ObjectAccessExpression : Expression, IPrintableExpression,
         INavigation navigation,
         Expression accessExpression,
         bool required)
+        : this(navigation, accessExpression, required,
+            navigation.TargetEntityType.GetContainingElementName() ??
+            throw new InvalidOperationException(
+                $"Navigation '{navigation.DeclaringEntityType.DisplayName()}.{navigation.Name}' doesn't point to an embedded entity."))
     {
-        Name = navigation.TargetEntityType.GetContainingElementName() ??
-               throw new InvalidOperationException(
-                   $"Navigation '{navigation.DeclaringEntityType.DisplayName()}.{navigation.Name}' doesn't point to an embedded entity.");
+    }
 
+    /// <summary>
+    /// Create a <see cref="ObjectAccessExpression"/> with an explicit field name.
+    /// Used for cross-collection $lookup results.
+    /// </summary>
+    /// <param name="navigation">The <see cref="INavigation"/> this object access relates to.</param>
+    /// <param name="accessExpression">The <see cref="Expression"/> of the parent containing the object.</param>
+    /// <param name="required">Whether this object is required.</param>
+    /// <param name="name">The explicit field name to access in the document.</param>
+    public ObjectAccessExpression(
+        INavigation navigation,
+        Expression accessExpression,
+        bool required,
+        string name)
+    {
+        Name = name;
         Navigation = navigation;
         AccessExpression = accessExpression;
         Required = required;
     }
+
+    /// <summary>
+    /// Create a <see cref="ObjectAccessExpression"/> for a cross-collection join result
+    /// where no navigation is defined (explicit Join).
+    /// </summary>
+    public ObjectAccessExpression(
+        IEntityType entityType,
+        Expression accessExpression,
+        bool required,
+        string name)
+    {
+        Name = name;
+        EntityType = entityType;
+        AccessExpression = accessExpression;
+        Required = required;
+    }
+
+    internal IEntityType? EntityType { get; }
 
     /// <inheritdoc />
     public override ExpressionType NodeType
@@ -46,11 +81,11 @@ internal sealed class ObjectAccessExpression : Expression, IPrintableExpression,
 
     /// <inheritdoc />
     public override Type Type
-        => Navigation.ClrType;
+        => Navigation?.ClrType ?? EntityType!.ClrType;
 
     public string Name { get; }
 
-    public INavigation Navigation { get; }
+    public INavigation? Navigation { get; }
 
     public Expression AccessExpression { get; }
 
@@ -62,7 +97,9 @@ internal sealed class ObjectAccessExpression : Expression, IPrintableExpression,
 
     public ObjectAccessExpression Update(Expression outerExpression)
         => outerExpression != AccessExpression
-            ? new ObjectAccessExpression(Navigation, outerExpression, Required)
+            ? (Navigation != null
+                ? new ObjectAccessExpression(Navigation, outerExpression, Required)
+                : new ObjectAccessExpression(EntityType!, outerExpression, Required, Name))
             : this;
 
     void IPrintableExpression.Print(ExpressionPrinter expressionPrinter)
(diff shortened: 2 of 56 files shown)


src/MongoDB.EntityFrameworkCore/Query/SerializerOverrideCollection.cs
```diff
@@ -0,0 +1,375 @@
...
+/// bare-collection queryable the join translator accepts, while still carrying EF's element-name /
+/// discriminator / BSON-representation mappings on the inner side.
+/// </para>
+/// <para>
+/// All other members delegate to the wrapped collection. The driver only ever reads
+/// <see cref="CollectionNamespace"/> and <see cref="DocumentSerializer"/> from this instance during
+/// query translation; the remaining members are present only to satisfy the interface.
+/// </para>
+/// </remarks>
+internal sealed class SerializerOverrideCollection<TDocument> : IMongoCollection<TDocument>
```
ajcvickers, line 54: TL;DR of the DL;DR; needed because of a driver limitation. A change in the driver could remove the need for it.

TL;DR from Claude: Cross-collection Include/navigation translates to a driver Join/GroupJoin. EF's entity serializer (element names, discriminators,  BsonRepresentation) must be attached to the join's inner operand — but the driver's join translator only accepts a bare collection-backed queryable and rejects the normal .As(serializer) operator there. Since the driver reads the inner serializer from collection.DocumentSerializer,  SerializerOverrideCollection is a thin IMongoCollection<T> decorator that overrides only DocumentSerializer to return EF's serializer, smuggling EF's mappings onto the inner side without breaking the join translator.

> "smuggling EF's mappings onto the inner side" 😆 

Me: Can this be removed with a small change in the driver?

TL;DR from Claude: Teach the driver's JoinMethodToPipelineTranslator to accept an .As(serializer)-wrapped inner operand (a MethodCallExpression) instead of only bare collection-backed queryable (ConstantExpression), and use that supplied serializer as the inner pipeline-input serializer — i.e. resolve the inner serializer from .As(...) the same way the outer side already does, rather than only from collection.DocumentSerializer. With that, the provider could call .As(...) on both sides and drop SerializerOverrideCollection.

src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoQueryExpression.cs
```diff
@@ -87,6 +89,108 @@ public Expression GetMappedProjection(ProjectionMember projectionMember)
...
 
+    /// <summary>
+    /// Pending $lookup stages for cross-collection collection Include operations, ordered so that a
+    /// transitive lookup (whose <see cref="LookupExpression.LocalField"/> matches against an
+    /// already-unwound intermediate lookup's <see cref="LookupExpression.As"/> field, e.g.
+    /// <c>_lookup_Order.CustomerID</c>) is emitted AFTER the lookup it depends on. Joins can be
+    /// registered in an order that doesn't respect this dependency, so we sort here.
+    /// </summary>
+    public IReadOnlyList<LookupExpression> PendingLookups
+        => OrderLookupsByDependency(_pendingLookups);
```
ajcvickers: Should be a method.

TL;DR²: Make PendingLookups a method — it sorts and allocates a new list each call (so a property lies about its cost), caching isn't worth it, and it's internal so renaming is free.

src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
```diff
@@ -30,27 +30,62 @@ public ObjectAccessExpression(
...
         Navigation = navigation;
         AccessExpression = accessExpression;
         Required = required;
     }
 
+    /// <summary>
+    /// Create a <see cref="ObjectAccessExpression"/> for a cross-collection join result
+    /// where no navigation is defined (explicit Join).
+    /// </summary>
+    public ObjectAccessExpression(
```
ajcvickers: Note: update docs here to be consistent.

src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
```diff
@@ -30,27 +30,62 @@ public ObjectAccessExpression(
...
+        IEntityType entityType,
+        Expression accessExpression,
+        bool required,
+        string name)
+    {
+        Name = name;
```
ajcvickers: Creating a work item to scan the code for correct use of strings verses strong type for field/property/element names.

src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
```diff
@@ -17,7 +17,7 @@ namespace MongoDB.EntityFrameworkCore.Query.Expressions;
 internal sealed class ObjectAccessExpression : Expression, IPrintableExpression, IAccessExpression
```
ajcvickers: Important: Equality/GetHashCode has a bug.

Claude's opinion on splitting: TL;DR: Don't split it just for the sake of it — the two shapes ("has Navigation" vs "EntityType-only") diverge in just two members (Type, Update), and splitting would break parity with the upstream Cosmos ObjectAccessExpression it's derived from. The real issue this surfaces is a bug: Equals/GetHashCode ignore both EntityType and Name, so distinct access nodes compare equal and can collapse incorrectly in projection dedup — fix that regardless. Only split if you want the "exactly one of Navigation/EntityType" invariant type-enforced and expect the two paths to diverge more over time (they haven't so far).

src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectArrayProjectionExpression.cs
ajcvickers: Can merge constructors, but call-sites will be slightly less clean due to resolving cases with missing optional parameters. Claude says to do it.

src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
ajcvickers: Decision: we will split

src/MongoDB.EntityFrameworkCore/Query/SerializerOverrideCollection.cs
ajcvickers: Fill an issue in the driver.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -141,8 +142,18 @@ protected override Expression VisitExtension(Expression extensionExpression)
                     var projectionBindingExpression =
                         (ProjectionBindingExpression)structuralTypeShaperExpression.ValueBufferExpression;
 
-                    var entityProjection = (EntityProjectionExpression)_queryExpression.GetMappedProjection(
-                        projectionBindingExpression.ProjectionMember);
+                    EntityProjectionExpression entityProjection;
```
ajcvickers, line 145: Are we suddenly using ordinal indexes? Check this.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -157,12 +168,67 @@ protected override Expression VisitExtension(Expression extensionExpression)
...
+                            // navigations were chained (e.g. OrderDetail.Order.Customer.Orders), the
+                            // reference chain is emitted as a series of root-level $lookup+$unwind
+                            // stages aliased "_lookup_<Nav>" rather than the driver's _outer/_inner
+                            // shape. A trailing collection Include whose declaring entity is one of
+                            // those unwound intermediates must match against that intermediate's
+                            // sub-document, so its $lookup localField needs the "_lookup_<Nav>." prefix.
+                            // The output "as" is nested under the same intermediate sub-document because
+                            // the shaper reads the collection array relative to the intermediate's
+                            // ParentAccessExpression (i.e. "_lookup_<Nav>._lookup_<Collection>").
+                            var declaringType = includableNavigation.DeclaringEntityType;
```
ajcvickers, line 215: Will two uses of the same navigation in a projection with different filters work corrcectly?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -157,12 +168,67 @@ protected override Expression VisitExtension(Expression extensionExpression)
...
+                            var intermediateLookup = _queryExpression.PendingLookups.FirstOrDefault(
```
ajcvickers: Make sure that intermediateLookup is uniquely named. Add test.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -521,6 +607,827 @@ protected override ElementInit VisitElementInit(ElementInit elementInit)
...
+
+    /// <summary>
+    /// Detects and binds a projected cross-collection collection-navigation Count such as
+    /// <c>select new { ..., TotalOrders = c.Orders.Count }</c>. EF Core lowers the count to
+    /// <c>Queryable.Count(Queryable.Where(DbSet&lt;Target&gt;(), joinPredicate))</c> (no predicate on the
+    /// Count). This registers a dedicated <c>_lookup_&lt;Nav&gt;</c> $lookup (flagged to be injected right
+    /// after the root collection source) and binds the count as a scalar projection. The EF-to-driver
+    /// translator rewrites the same subtree into a server-side <c>{ $size: "$_lookup_&lt;Nav&gt;" }</c>.
+    /// </summary>
+    private bool TryBindProjectedCollectionNavigationCount(
```
ajcvickers: Why no predicate versions? Also, what if the source already has a filter/where?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -521,6 +607,827 @@ protected override ElementInit VisitElementInit(ElementInit elementInit)
...
+        // Find the outer entity shaper referenced by the join predicate and the collection navigation
+        // (off the outer entity) whose target is this DbSet.
+        var outerShaper = FindOuterShaper(whereCall.Arguments[1]);
+        if (outerShaper?.StructuralType is not IEntityType outerEntityType)
+        {
+            return false;
+        }
+
+        INavigation navigation = null;
+        foreach (var candidate in outerEntityType.GetNavigations())
```
ajcvickers: Make sure it doesn't just use the entity type to determine a relationship end.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -521,6 +607,827 @@ protected override ElementInit VisitElementInit(ElementInit elementInit)
...
+            if (candidate.IsCollection
+                && !candidate.IsEmbedded()
+                && candidate.TargetEntityType == targetEntityType)
+            {
+                navigation = candidate;
+                break;
+            }
+        }
+
+        if (navigation == null)
```
ajcvickers: Check unidirectional relationships where the navigation doesn't exist, but the relationship does.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -521,6 +607,827 @@ protected override ElementInit VisitElementInit(ElementInit elementInit)
...
+            if (found == null && shaper.ValueBufferExpression is ProjectionBindingExpression)
+            {
+                found = shaper;
+            }
+        });
+        finder.Visit(predicate);
+        return found;
+    }
+
+    private sealed class ShaperFinder(Action<StructuralTypeShaperExpression> onFound) : ExpressionVisitor
```
ajcvickers: Why does this need a visitor, and what does the onFound do?

ajcvickers: Comment this better.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -521,6 +607,827 @@ protected override ElementInit VisitElementInit(ElementInit elementInit)
...
+        Expression innerShaper,
+        ObjectArrayProjectionExpression parentArrayProjection)
+    {
+        var selectorArg = selectCall.Arguments[1];
+        while (selectorArg is UnaryExpression { NodeType: ExpressionType.Quote } quote)
+        {
+            selectorArg = quote.Operand;
+        }
+
+        if (selectorArg is not LambdaExpression lambda)
```
ajcvickers: When is it not a lambda?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -353,7 +350,195 @@ protected override QueryableMethodTranslatingExpressionVisitor CreateSubqueryVis
...
+        if (navigation == null && fkPropertyName != null)
+        {
+            foreach (var priorInnerEntityType in outerQueryExpression.InnerCollections.Keys)
+            {
+                if (priorInnerEntityType == innerEntityType)
+                {
+                    continue;
+                }
+
+                var candidate = priorInnerEntityType.GetNavigations()
```
ajcvickers, line 436: Should this not be all?

ajcvickers: Double check this.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -353,7 +350,195 @@ protected override QueryableMethodTranslatingExpressionVisitor CreateSubqueryVis
...
+        // never contradicts the emitted pipeline.
+        //
+        // Each cross-collection projection carries its OWNING navigation and a stable
+        // "_lookup_<Navigation>" alias. The shaper derives the field it reads from that navigation
+        // plus the computed UsesDriverJoinFields flag (driver-native => "_inner"; flat => the
+        // "_lookup_<Navigation>" alias), so the projection is never retroactively rewritten.
+        var isSecondOrLaterJoin = outerQueryExpression.InnerCollections.Count > 1;
+        if (isSecondOrLaterJoin)
+        {
+            // Flatten: register a forced-unwind $lookup for THIS join...
```
ajcvickers, line 464: If all are flattened to root, how are the associations between second and third preserved for fixup?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -353,7 +350,195 @@ protected override QueryableMethodTranslatingExpressionVisitor CreateSubqueryVis
...
+
+        return structuralShaper.Update(
+            new ProjectionBindingExpression(outerQueryExpression, projectionIndex, typeof(ValueBuffer)));
+    }
+
+    /// <summary>
+    /// Extract the property name from a key selector lambda (e.g., o => o.CustomerId -> "CustomerId").
+    /// Handles EF.Property calls and direct member access.
+    /// </summary>
+    private static string? ExtractPropertyName(LambdaExpression keySelector)
```
ajcvickers: Composite keys?
DRY?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -436,9 +1254,218 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+    /// <summary>
+    /// A projected collection-navigation count is materialized by a single <c>$lookup</c> injected right
+    /// after the root source. Under a set operation (e.g. <c>Union</c>) where more than one branch reads
+    /// the looked-up collection, the non-leading branch becomes a <c>$unionWith</c> sub-pipeline that does
+    /// not see that root-level <c>$lookup</c>, so its server-side <c>{ $size: "$_lookup_&lt;Nav&gt;" }</c>
+    /// would fail at runtime with "argument to $size must be an array". Detect that shape and fail
+    /// translation cleanly (an <see cref="InvalidOperationException"/>) instead of emitting a pipeline that
+    /// crashes on the server.
+    /// </summary>
+    private void GuardAgainstMultiBranchNavigationCount(Expression? efQueryExpression)
```
ajcvickers, line 607: What is this limitation practically? Can we do a select projection with two counts? What would you ideally generate as MQL?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -46,44 +47,708 @@ internal sealed class MongoEFToLinqTranslatingExpressionVisitor : System.Linq.Ex
...
+    /// all handled.
+    ///
+    /// <paramref name="convertExplicitJoins"/> controls whether an explicit user <c>Queryable.Join</c> also
+    /// gets a <see cref="LeftJoinResult{TOuter,TInner}"/> result selector. The shaped (entity-materializing)
+    /// path needs this — its client-side shaper reads <c>_outer</c>/<c>_inner</c> from the joined document.
+    /// The projected path must NOT convert explicit Joins: their result selector is the user's projection and
+    /// has to be emitted verbatim. (A <c>LeftJoin</c> is always converted regardless, since the driver has no
+    /// LeftJoin translator.)
+    /// </summary>
+    private Expression RewriteLeftJoins(Expression expression, bool convertExplicitJoins)
```
ajcvickers: Can we separate out all the LeftJoin so that it can easily be replaced by driver code.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -46,44 +47,708 @@ internal sealed class MongoEFToLinqTranslatingExpressionVisitor : System.Linq.Ex
...
+        var serializerType = typeof(BsonClassMapSerializer<>).MakeGenericType(resultType);
+        return (IBsonSerializer)Activator.CreateInstance(serializerType, classMap)!;
+    }
+
+    /// <summary>
+    /// Resolves the dotted BSON field path for a simple <c>EF.Property(p, "Name")</c> / member-access join
+    /// key selector body. Returns <see langword="null"/> for shapes we don't handle (composite/anonymous
+    /// keys, conversions, computed keys).
+    /// </summary>
+    private static string? TryGetKeyFieldPath(Expression keyBody, IEntityType entityType)
```
ajcvickers: DRY?

src/MongoDB.EntityFrameworkCore/Query/Expressions/ObjectAccessExpression.cs
ajcvickers: I traced every MQL-producing path. The provider is genuinely disciplined about this — I found no case where a raw C# name leaks into MQL where a mapped element name was required. The design already follows exactly the pattern you described (strong type carried as far as possible, raw string only at the  end):

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
ajcvickers: Critical: this limitation will be removed by driver LeftJoin.

Claude TL;DR: It guards against a set operation (Union/Concat/Except/Intersect) where 2+ branches each project a collection-navigation count (e.g. c.Orders.Count).
  
Counts are translated as a root-level $lookup (InjectAfterRoot) that materializes _lookup_<Nav>, plus a server-side { $size: "$_lookup_<Nav>" }. But only the leading branch gets that root $lookup; other branches become $unionWith sub-pipelines that can't see it, so $size hits a missing field → runtime  "argument to $size must be an array." The guard detects this shape and throws a clean EF TranslationFailed instead of shipping a crashing pipeline. (One of the workarounds native LeftJoin/EF-317 would remove.)

Full answer:

It's guarding against a specific shape that the projected collection-count workaround can't express: a set operation (Union/Concat/Except/Intersect) where two or more branches each project a collection-navigation count (e.g. c.Orders.Count).

  Here's the exact mechanism and why it breaks:

  How a nav-count is normally translated. The provider doesn't emit a correlated subquery for c.Orders.Count. Instead it:
  1. registers one $lookup (flagged LookupExpression.InjectAfterRoot) that materializes the related collection into a root-level array field _lookup_<Nav>, injected right after the root collection source, and
  2. rewrites c.Orders.Count → server-side { $size: "$_lookup_<Nav>" } (TryRewriteCollectionNavigationCount), which reads that materialized array.
  
  The whole thing rests on one assumption: the _lookup_<Nav> array is present at the document root wherever the $size runs.

  Why a multi-branch set operation violates it. A set operation compiles to MongoDB's $unionWith (and friends). Only the leading branch is the outer pipeline that receives the injected root-level $lookup. Every non-leading branch becomes a $unionWith sub-pipeline — a separate pipeline that does not inherit
  the outer pipeline's stages, so _lookup_<Nav> simply doesn't exist there. The rewritten { $size: "$_lookup_<Nav>" } in that branch then dereferences a missing field, and the server fails at runtime with "argument to $size must be an array."

  So the limitation is architectural: the workaround can inject the count's $lookup at exactly one root position, but a multi-branch set op needs the array materialized independently inside each branch's sub-pipeline, which the current $lookup+$size scheme doesn't do.

  What the guard does. Rather than emit a pipeline that crashes cryptically on the server, GuardAgainstMultiBranchNavigationCount detects the shape — a set-operation method call with ≥2 operands that each contain the Count(Where(EntityQueryRootExpression, fk)) subtree — and throws the standard EF
  CoreStrings.TranslationFailed InvalidOperationException instead. It only bothers checking when at least one InjectAfterRoot lookup is actually pending.

  It pairs with the rewrite's own guard (TryRewriteCollectionNavigationCount only fires for a single InjectAfterRoot lookup): single-branch count → rewritten and works; multi-branch count → caught here → clean translation failure.

  This is squarely one of the workaround limitations that native driver LeftJoin (EF-317) would remove — it lives in the $lookup-emission code now quarantined in MongoEFToLinqTranslatingExpressionVisitor.LeftJoin.cs.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: TL;DR: A combination. A ProjectionBindingExpression is keyed by either a named ProjectionMember (built from a member-path stack during projection traversal) or an ordinal Index (used directly for cross-collection joins and after ApplyProjection flattens the mapping). GetProjectionIndex normalizes both to a single ordinal index into the Projection list — member-keyed ones resolve indirectly via _projectionMapping to a Constant<int>.

Separately, the actual document read underneath is string-based (projection Alias / element names / _lookup_<Nav> fields) — a different layer from the binding identity.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: No, but should be fixed by driver LeftJoin.

Claude:

Short answer: No — the code cannot represent the same navigation collection with two different filters in one query. Here's why, in layers.
  
  Architectural reason: one $lookup slot per navigation

  - The lookup's output field is As = _lookup_<navigation.Name> — keyed only on the navigation name, with no filter component (LookupExpression.GetLookupAlias).
  - AddLookup dedupes by that alias and silently skips any second registration for the same navigation:
  if (!_pendingLookups.Any(l => l.As == lookup.As)) _pendingLookups.Add(lookup);
  - A LookupExpression carries a single PipelineStages list. So a navigation can hold at most one filtered materialization per query; both read sites resolve to the same _lookup_<Nav> array. There is no per-filter disambiguation and no guard that detects a conflicting second registration — the first one
  silently wins.

  What actually happens for the shapes that would need it

  In practice you don't usually hit silent wrong-results, because the shapes that would require two filters on one nav are rejected earlier:

  1. Filtered Include with a Where predicate (Include(c => c.Orders.Where(o => o.Total > 100))) — ExtractFilteredIncludePipeline only drops the synthetic FK-correlation predicate; any user Where (or HasQueryFilter) it can't fold throws TranslationFailed. So predicate-filtered includes aren't supported at
  all, regardless of count.
  2. Filtered projected collections in a Select (c => new { Big = c.Orders.Where(p).ToList(), Small = c.Orders.Where(q).ToList() }) — TryBindProjectedCollectionNavigation accepts only the bare FK-join Where and returns false on any extra Where/OrderBy/Skip/Take. So these never route through the $lookup
  path here.
  3. OrderBy/Skip/Take-shaped filtered includes are supported (→ $sort/$skip/$limit in the lookup sub-pipeline), but only one per navigation — and EF Core itself rejects conflicting duplicate filtered includes on the same navigation, so two-different-on-one-nav doesn't reach the provider.

  Bottom line

  Two unfiltered references to the same navigation dedupe correctly (identical lookups). Two different filters on the same navigation are not supported: predicate filters fail translation upfront; ordering/paging variants can't be registered twice for one nav and would be silently deduped if they ever got
  there. The single-$lookup-per-navigation keying is exactly the kind of constraint native driver LeftJoin (EF-317) would lift.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: It's fine in of itself, but did reveal another issue:

TL;DR: intermediateLookup was matched by entity type via FirstOrDefault, so when two reference/forceUnwind lookups targeted the same declaringType (two nav props to the same type, or a self-ref chain) it silently picked one and could prefix the $lookup with the wrong _lookup_<Nav>. path → wrong results. I replaced it with a .Where(...).ToList(): >1 candidate now throws TranslationFailed, exactly 1 behaves as before, 0 still skips. The ambiguity fails loudly instead of silently. Builds clean on EF10.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: TL;DR: Both restrictions exist because the count becomes `{ $size: "$_lookup_<Nav>" }` over the fully-materialized lookup array, and $size counts every element with no filtering.
  
- Predicate excluded: only 1-arg Count/LongCount is accepted. Count(o => predicate) would need to count only matching elements, but $size returns the whole array length — accepting it would silently ignore the predicate and return the total. So it's rejected.
- No filter on source: the source must be exactly Where(EntityQueryRootExpression, fkEquality) — the bare DbSet with only the synthetic FK-join predicate. Any extra Where/OrderBy/Skip/Take makes Arguments[0] another Where instead of the root, the pattern fails, and it returns false — again because $size over the unfiltered array can't honor it.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: Fixed.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: Fixed.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -521,6 +607,827 @@ protected override ElementInit VisitElementInit(ElementInit elementInit)
...
+        _projectionMapping[projectionMember] = methodCallExpression;
+
+        result = new ProjectionBindingExpression(_queryExpression, projectionMember, methodCallExpression.Type);
+        return true;
+    }
+
+    /// <summary>
+    /// Unwrap a (possibly quoted) lambda selector, returning its body and single parameter.
+    /// </summary>
+    private static Expression UnwrapLambdaBody(Expression selector, out ParameterExpression parameter)
```
ajcvickers: DRY?

ajcvickers: What was consolidated

  ExpressionExtensionMethods (the single home) gained/unified four unwrap helpers:
  - RemoveConvert (already existed) — recursive strip of all Convert/ConvertChecked.
  - RemoveObjectConvert — moved here from MongoEFToLinqTranslatingExpressionVisitor (distinct semantics: strips only a single, object-typed boxing convert; documented to contrast with RemoveConvert).
  - UnwrapQuote — new; the canonical Quote-stripping loop.
  - UnwrapLambdaFromQuote — refactored to delegate to UnwrapQuote (one quote-stripping implementation now).
  - TryGetSimplePropertyName — new; the member-access / EF.Property(...) name extractor, built on RemoveConvert + IsEFPropertyMethod.

  Duplicates removed:
  - ExtractPropertyName (MongoQueryableMethodTranslatingExpressionVisitor) and TryGetPropertyNameFromKeySelector (LeftJoin) — two copies of the same name-extraction logic — both deleted, callers now use TryGetSimplePropertyName.
  - The private RemoveObjectConvert in the EF→LINQ visitor — deleted, its 5 call sites now use the shared extension.
  - 5 inline Convert/ConvertChecked unwrap loops (in TryResolveFieldAccess, TryResolveFieldAccessSource, my own RootsAtParameter, plus the two extractors) → RemoveConvert().
  - 4 inline Quote unwrap loops (UnwrapLambdaBody — the example you cited — plus three selector-unwrap sites) → UnwrapQuote().
  - GetSortField rewritten from a 25-line quote+convert+member walk to 3 lines via UnwrapLambdaFromQuote().Body.TryGetSimplePropertyName().

  Where I held back (judgment)

  - Did not merge RemoveObjectConvert into RemoveConvert — they have genuinely different semantics (single object-boxing strip vs. recursive strip-all). Merging would have silently changed behavior at the boxing-aware call sites. I centralized it instead and documented the distinction.
  - Left UnwrapLambdaBody as its own method (it returns body + out parameter, a distinct shape) — but routed its quote-stripping through the shared UnwrapQuote.
  - Left the inline is UnaryExpression { NodeType: Quote, Operand: LambdaExpression } pattern-matches in the LeftJoin rewriter (lines 172/273) — those destructure a known shape inline rather than loop-unwrap, and forcing them through a helper would be less clear, not more.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: Essentially, never.

Refactored.

src/MongoDB.EntityFrameworkCore/Query/SerializerOverrideCollection.cs
ajcvickers: Issues filed.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
ajcvickers: Tracked as part of EF-317

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
ajcvickers: See comments in EF-317

src/MongoDB.EntityFrameworkCore/Storage/BsonBinding.cs
```diff
@@ -204,6 +204,15 @@ private static readonly MethodInfo GetElementValueMethodInfo
...
     internal static T? GetPropertyValue<T>(BsonDocument document, IReadOnlyProperty property)
     {
+        // A null parent document means the owning entity is absent entirely (e.g. an optional
+        // cross-collection reference nested inside a collection Include whose $lookup matched no
+        // document). Treat every property as absent so the entity materializer's null-key check
+        // produces a null entity rather than dereferencing the missing document.
+        if (document == null)
+        {
+            return default;
+        }
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/UnsupportedQueryTests.cs
```diff
@@ -30,24 +30,19 @@ public sealed class UnsupportedQueriesTests(ReadOnlySampleGuidesFixture database
...
-    public void Join_cannot_be_translated()
+    public void Join_can_be_translated()
     {
-        var ex = Assert.Throws<InvalidOperationException>(() =>
-            _db.Planets.Join(_db.Moons, p => p._id, m => m.planetId, (p, m) => new { p, m }).ToList());
-        Assert.Contains(".Join(", ex.Message);
-        Assert.Contains(" could not be translated", ex.Message);
+        var result = _db.Planets.Join(_db.Moons, p => p._id, m => m.planetId, (p, m) => new { p, m }).ToList();
+        Assert.NotNull(result);
     }
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

tests/MongoDB.EntityFrameworkCore.FunctionalTests/Query/UnsupportedQueryTests.cs
```diff
@@ -30,24 +30,19 @@ public sealed class UnsupportedQueriesTests(ReadOnlySampleGuidesFixture database
...
+    public void LeftJoin_can_be_translated()
     {
-        var ex = Assert.Throws<InvalidOperationException>(
-            () => _db.Planets.LeftJoin(_db.Moons, p => p._id, m => m.planetId, (p, m) => new {p, m}).ToList());
-
-        Assert.Contains(".LeftJoin(", ex.Message);
-        Assert.Contains(" could not be translated", ex.Message);
+        var result = _db.Planets.LeftJoin(_db.Moons, p => p._id, m => m.planetId, (p, m) => new {p, m}).ToList();
+        Assert.NotNull(result);
     }
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: Make sure we are tracking this.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
ajcvickers: Track this.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
ajcvickers: @damieng to look into

ajcvickers (commented):
LGTM

damieng (approved):
Arthur and I went over this in the course of two days pairing on the review. All issues and questions have been addressed in the various updates to this PR since its initial submission.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
damieng: Am happy with this now.  Reminder: The index is EF's ordinal into the query's projection list (how the shaper points at a slot); the name/alias is the BSON element name that says where the value actually lives in the document - it resolves index to ProjectionExpression to alias, because the name only becomes knowable after the index dereference.

PR #310, opened 2026-06-08
EF-316: Harden Atlas TestContainers startup and readiness probe

ajcvickers:
Port the container-readiness hardening from mongo-mevd-provider#2 (which
itself was ported from this repo's test infra):

- MongoDbAtlasBuilder.WaitIndicateReadiness: scope the probe's catch to
  MongoException/TimeoutException instead of a bare catch, so unexpected
  exceptions propagate rather than being silently retried forever; add
  ConfigureAwait(false).
- TestContainersTestServer.InitializeAsync: bound StartAsync with a 5-minute
  CancellationTokenSource so a stuck readiness probe (e.g. the Atlas Search
  Index Management service never coming up) fails fast instead of hanging the
  run; dispose the container on a failed start and dispose any leftover
  container before allocating a new one.


ajcvickers:
@damieng The GitHub action passed 10 times in a row, so I think this is ready to merge--it fixes a few flaky things running the Atlas tests.

damieng (approved):
LGTM

PR #311, opened 2026-06-11
EF-107: Bulk ExecuteUpdate/ExecuteDelete

ajcvickers:
- Implements server-side bulk `ExecuteDelete`/`ExecuteUpdate` (+ async) for EF9/EF10
- `Where`-scoped ops run as one atomic `deleteMany`/`updateMany`
- `OrderBy`/`ThenBy`/`Skip`/`Take`/`Distinct` sources run as a transactional two-phase op: collect target `_id`s, then act via `{ _id: { $in: … } }`.
- Two-phase uses the ambient transaction or auto-starts one
- Out-of-subset shapes surface EF's canonical "could not be translated" error.
- Implements EF Core's `NorthwindBulkUpdates` conformance suite

## Execution strategy by source shape

Single round-trip — one atomic command, no transaction:
- `Where`-scoped (any number of `Where` predicates), single collection. Runs as `deleteMany(filter)` / `updateMany(filter, <$set|pipeline>)`. Atomic at the server; no transaction started.

Two-phase — transactional (`_id` projection → `$in` write):
- `Where` combined with `OrderBy` / `OrderByDescending` / `ThenBy` / `ThenByDescending` / `Skip` / `Take` / `Distinct`.
- Phase 1 reads the matched documents and collects their stored `_id` (scalar or composite sub-document); phase 2 runs `deleteMany`/`updateMany` filtered by `{ _id: { $in: ids } }`.
- Both phases share one transaction: the ambient one if open (provider does not commit it), otherwise an auto-started transaction (committed on success, aborted on failure). Requires a replica set / sharded cluster; `AutoTransactionBehavior.Never` throws and asks for an explicit transaction.

Not supported — rejected at translation:
- Joins / correlated subqueries (`Join`/`LeftJoin`/`RightJoin`/`GroupJoin`, cross/outer apply), `GroupBy`, `SelectMany`, set operations (`Except`/`Intersect`; `Concat`/`Union` blocked by the `$unionWith`-in-transaction server restriction), cross-document navigation predicates, non-entity projection sources, and multiple-collection updates.
- All surface as `InvalidOperationException` ("could not be translated"). Unblocking the join/group/set families needs read-side translation that doesn't exist yet (ultimately a driver `$lookup`/`$group` + `$merge` capability). On EF8, all bulk operations are rejected.

## Details
- New internal marker node `MongoNonQueryExpression` — Delete/Update kind, ordered setters, `SingleCommand`/`TwoPhase` strategy.
- Translation (`MongoQueryableMethodTranslatingExpressionVisitor`): `ClassifyBulkSource` walks the captured chain to the strategy or the canonical rejection; `BuildSetter` resolves `SetProperty` / `EF.Property` targets to root scalar properties and classifies constant vs self-referencing; empty-setter rejected.
- Execution (`MongoShapedQueryCompilingExpressionVisitor`): single-command `BuildFilter` + `deleteMany`/`updateMany`; two-phase `SelectTargetIds` (reuses the read-path translation to yield raw documents and reads `_id`) inside a transaction scope, then the `{ _id: $in }` write; `BuildUpdate` shared (`$set`, or pipeline `$set` when any setter is self-referencing).
- Diagnostics: `Executing/ExecutedBulkDelete` and `Executing/ExecutedBulkUpdate`; the two-phase target count rides on a new additive `MongoBulkWriteEventData.TargetCount` (Delete/ModifyCount stay 0 on the pre-execution event).
- Docs: design + implementation plan under `docs/superpowers/`; `docs/failing-spec-tests.md` (EF-X016 with remaining shapes grouped by root cause); README "What is supported".

## Testing
- Functional `ExecuteDeleteTests` / `ExecuteUpdateTests`: single-command, two-phase ordering/paging/Distinct, rollback inside a user transaction, empty target, `AutoTransactionBehavior.Never`, self-referencing setter, and diagnostics events.
- Spec `NorthwindBulkUpdatesMongoTest`: supported shapes call `base` and assert affected count + before/after state; out-of-subset shapes run and assert their actual failure mode (tagged per `docs/failing-spec-tests.md`); 18 ordering/paging/Distinct cases promoted to green (incl. composite-key `OrderDetail`).
- Full suite green on EF8 / EF9 / EF10 (0 failures; only environment-gated Queryable-Encryption skips).


src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -281,6 +334,739 @@ private static QueryingEnumerable<BsonDocument, TResult> ExecuteShapedQuery<TSou
...
+        {
+            return body();
+        }
+
+        if (database.AutoTransactionBehavior == AutoTransactionBehavior.Never)
+        {
+            throw NoAutoTransactionError();
+        }
+
+        var transaction = database.BeginTransaction();
```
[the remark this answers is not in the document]
ajcvickers: Moved begin inside the try in both helpers; null-safe rollback/dispose so a begin-time failure still cleans up correctly

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -281,6 +334,739 @@ private static QueryingEnumerable<BsonDocument, TResult> ExecuteShapedQuery<TSou
...
+            return await body().ConfigureAwait(false);
+        }
+
+        if (database.AutoTransactionBehavior == AutoTransactionBehavior.Never)
+        {
+            throw NoAutoTransactionError();
+        }
+
+        var cancellationToken = queryContext.CancellationToken;
+        var transaction = await database.BeginTransactionAsync(cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
ajcvickers: Moved begin inside the try in both helpers; null-safe rollback/dispose so a begin-time failure still cleans up correctly

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -281,6 +334,739 @@ private static QueryingEnumerable<BsonDocument, TResult> ExecuteShapedQuery<TSou
...
+            + "as a two-phase operation requiring a transaction. The current MongoDB deployment does not support "
+            + "multi-document transactions (a replica set or sharded cluster is required).", inner);
+
+    // Multi-document transactions are rejected on standalone deployments; the driver surfaces this as a
+    // MongoCommandException (code 20 / IllegalOperation) or a "Transaction numbers are only allowed on a
+    // replica set member or mongos" message. Match conservatively so unrelated failures propagate untouched.
+    private static bool IsTransactionsUnsupported(Exception exception)
+        => exception is MongoCommandException { Code: 20 }
+           || (exception is MongoException && exception.Message.Contains("Transaction numbers are only allowed", StringComparison.Ordinal));
+
```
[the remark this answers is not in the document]
ajcvickers: Added a third match arm for NotSupportedException containing "does not support transactions" (covers both the standalone-config and version branches of MongoTransaction.Start)

tests/MongoDB.EntityFrameworkCore.SpecificationTests/MongoComplianceTest.cs
```diff
@@ -115,11 +115,20 @@ public class MongoComplianceTest : ComplianceTestBase
...
         typeof(Microsoft.EntityFrameworkCore.BulkUpdates.BulkUpdatesTestBase<>),
         typeof(Microsoft.EntityFrameworkCore.BulkUpdates.FiltersInheritanceBulkUpdatesTestBase<>),
         typeof(Microsoft.EntityFrameworkCore.BulkUpdates.InheritanceBulkUpdatesTestBase<>),
         typeof(Microsoft.EntityFrameworkCore.BulkUpdates.NonSharedModelBulkUpdatesTestBase),
-        typeof(Microsoft.EntityFrameworkCore.BulkUpdates.NorthwindBulkUpdatesTestBase<>),
+        // NorthwindBulkUpdatesTestBase is now implemented by NorthwindBulkUpdatesMongoTest: the supported
+        // single-collection Where-scoped scalar cases, as well as OrderBy/Skip/Take/Distinct-scoped cases
+        // (executed via the two-phase _id-projection path), call base and pass; everything outside that
+        // subset (joins, set ops, GroupBy, SelectMany, navigations, non-entity projections,
+        // multiple-collection updates) is skipped there with a scope reason.
```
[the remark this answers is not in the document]
ajcvickers: Reworded to "not skipped — each runs and asserts its actual failure mode"

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoBulkWriteEventData.cs
```diff
@@ -46,13 +46,41 @@ public MongoBulkWriteEventData(
...
+    /// </param>
+    public MongoBulkWriteEventData(
+        EventDefinitionBase eventDefinition,
+        Func<EventDefinitionBase, EventData, string> messageGenerator,
+        TimeSpan elapsed,
+        CollectionNamespace collectionNamespace,
+        long insertCount,
+        long deleteCount,
+        long modifyCount,
+        int? targetCount)
```
damieng: We should introduce either two new eventdata classes - one for the bulk delete operations and one for bulk update operations that has the new target count (maybe also make it a long for consistency) as otherwise this is getting muddy as it's curently used for MongoDB "bulk" operations that are really just batches inside of SaveChanges and not EF bulk operations at all.

docs/agentplans/2026-06-11/bulk-two-phase-execute-update-delete/bulk-two-phase-execute-update-delete.design.md
```diff
@@ -0,0 +1,214 @@
+# Two-phase `ExecuteUpdate` / `ExecuteDelete` for ordered / paged / distinct sources — Design
```
damieng, line 1: I like the idea of keeping plans around for a while but I'm not sure the repo is the place for that. They're also going to get out of date quite quickly. Are they a historical artifact or part of ongoing documentation?  Perhaps we should attach them to the JIRA tickets instead.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoEventId.cs
```diff
@@ -226,4 +230,48 @@ private static EventId MakeDatabaseId(Id id)
...
+    /// <summary>
+    /// A bulk update (server-side <c>ExecuteUpdate</c>) has been executed.
+    /// </summary>
+    /// <remarks>
+    /// <para>This event is in the <see cref="DbLoggerCategory.Update" /> category.</para>
+    /// <para>
+    /// This event uses the <see cref="MongoBulkWriteEventData" /> payload when used with a <see cref="DiagnosticSource" />.
+    /// </para>
+    /// </remarks>
+    public static readonly EventId ExecutedBulkUpdate = MakeUpdateId(Id.ExecutedBulkUpdate);
```
damieng, line 287: Separate messages, I like this.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerUpdateExtensions.cs
```diff
@@ -22,6 +22,11 @@
 namespace MongoDB.EntityFrameworkCore.Diagnostics;
 
 using BulkWriteEventDefinition = EventDefinition<string, CollectionNamespace, long, long, long>;
+#if !EF8
+using BulkExecutingEventDefinition = EventDefinition<string, CollectionNamespace>;
+using BulkExecutingTwoPhaseEventDefinition = EventDefinition<string, CollectionNamespace, int>;
+using BulkExecutedEventDefinition = EventDefinition<string, CollectionNamespace, long>;
+#endif
```
damieng, line 29: Do we want to support EF9 for this? I thought there were some constraints on it prior to 10?

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerUpdateExtensions.cs
```diff
@@ -171,4 +176,338 @@ private static BulkWriteEventDefinition LogExecutedBulkWrite(IDiagnosticsLogger
...
+                static logger => new BulkExecutedEventDefinition(
+                    logger.Options,
+                    MongoEventId.ExecutedBulkUpdate,
+                    LogLevel.Information,
+                    "MongoEventId.ExecutedBulkUpdate",
+                    level => LoggerMessage.Define<string, CollectionNamespace, long>(
+                        level,
+                        MongoEventId.ExecutedBulkUpdate,
+                        "Executed Bulk Update ({elapsed} ms) Collection='{collectionNamespace}', Modified={modified}"))));
+#endif
```
damieng, line 504: I wonder if we should split these out by type into their own files now given there's x methods for each message?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -177,6 +185,248 @@ protected override ShapedQueryExpression CreateShapedQueryExpression(IEntityType
...
+        // unrelated invocation) must be rejected here rather than silently running a no-op updateMany.
+        if (parsed.Count == 0)
+        {
+            AddTranslationErrorDetails(
+                "An 'ExecuteUpdate' call must specify at least one 'SetProperty' invocation, "
+                + "to indicate the properties to be updated.");
+            throw new InvalidOperationException(
+                CoreStrings.NonQueryTranslationFailedWithDetails(
+                    mongoQueryExpression.CapturedExpression?.Print(), TranslationErrorDetails));
+        }
```
damieng, line 242: Do we want to support EF9 given this workaround/limitation? Do we need to document the differences?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -177,6 +185,248 @@ protected override ShapedQueryExpression CreateShapedQueryExpression(IEntityType
...
+                    break;
+
+                case nameof(Queryable.OrderBy):
+                case nameof(Queryable.OrderByDescending):
+                case nameof(Queryable.ThenBy):
+                case nameof(Queryable.ThenByDescending):
+                case nameof(Queryable.Skip):
+                case nameof(Queryable.Take):
+                case nameof(Queryable.Distinct):
+                    strategy = MongoNonQueryExpression.BulkStrategy.TwoPhase;
```
damieng, line 403: Are there complex queries within a where that might not translate in a single phase?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -57,6 +65,50 @@ public MongoShapedQueryCompilingExpressionVisitor(
...
+    {
+        var entityType = nonQueryExpression.SourceQuery.CollectionExpression.EntityType;
+
+        if (nonQueryExpression.Strategy == MongoNonQueryExpression.BulkStrategy.TwoPhase)
+        {
+            // Two-phase: phase 1 materializes matched entities and extracts their stored _id,
+            // phase 2 deletes/updates by { _id: { $in: [...] } } — both inside a transaction. Works for
+            // scalar and composite primary keys because we read the raw _id BsonValue from the serialized doc.
+            EnsureBulkKeyOrThrow(entityType, nonQueryExpression);
+            var twoPhaseExecutor = nonQueryExpression.Kind == MongoNonQueryExpression.OperationKind.Update
```
damieng: This and the pattern match below are a bit verbose/unclear/repetitive FWIW.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -311,6 +363,765 @@ private static QueryingEnumerable<BsonDocument, TResult> ExecuteShapedQuery<TSou
...
+    // grows retry/cursor concerns that belong in the wrapper.
+
+    // Two-phase bulk needs phase 1 (read) and phase 2 (write) to observe one snapshot, so both run inside a
+    // single transaction. If the user already opened one we join it (and never commit it — they own it);
+    // otherwise we auto-start one, commit on success, abort on failure. AutoTransactionBehavior.Never means
+    // "I manage transactions" — we refuse to auto-start and tell the caller to open one.
+    private static int RunInBulkTransaction(QueryContext queryContext, MongoNonQueryExpression nonQuery, Func<int> body)
+    {
+        var database = queryContext.Context.Database;
+        if (database.CurrentTransaction != null)
```
damieng: This really needs to go through our MongoTransaction class or we're going to lose the commit/rollback logging, the Guid for the transaction etc.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -311,6 +363,765 @@ private static QueryingEnumerable<BsonDocument, TResult> ExecuteShapedQuery<TSou
...
+                return 0;
+            }
+
+            var update = TranslateBulkOrThrow(nonQuery,
+                () => BuildUpdate<TSource>(queryContext, entityType, bsonSerializerFactory, nonQuery));
+            return UpdateByIds(queryContext, nonQuery, ids, update);
+        });
+
+    private static Task<int> ExecuteTwoPhaseUpdateAsync<TSource>(
+        QueryContext queryContext, IReadOnlyEntityType entityType, BsonSerializerFactory bsonSerializerFactory,
```
damieng: All this update logic in the QueryCompilingExpressionVisitor should be living somewhere else more update related.

damieng (changes requested):
I think we need to separate out the message event class and move the update/execution code out of the visitor ideally leveraging the existing transaction logic as much as possible.

docs/agentplans/2026-06-11/bulk-two-phase-execute-update-delete/bulk-two-phase-execute-update-delete.design.md
ajcvickers: Just talked to James, and we're pretty sure that we want to keep these in the GitHub repos going forward. However, I will change the file structure and make it clear what they are.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerUpdateExtensions.cs
ajcvickers: Bulk ExecuteUpdate/ExecuteDelete is fully supported and tested on EF9. There was an implementation change in EF10 that improves the user-experience of the API (user has to handle fewer expression trees) and means we need different code, but it didn't change the underlying functionality.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoBulkWriteEventData.cs
ajcvickers: Done.

src/MongoDB.EntityFrameworkCore/Diagnostics/MongoLoggerUpdateExtensions.cs
ajcvickers: We could, but it's all just boilerplate code. The relational version of this file has 3695 lines, and the core one has 3791 lines. Personally, I don't think this is too much.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
ajcvickers: EF10 validates "at least one SetProperty" before the call reaches the provider, so on EF9 we re-implement that one guard rather than silently running a no-op updateMany It's not a behavioral limitation. The underlying functionality is the same on EF9 and EF10.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoQueryableMethodTranslatingExpressionVisitor.cs
ajcvickers: Yes, but ClassifyBulkSource only inspects the operator chain. It doesn't inspect the predicate's contents. A Where whose predicate itself isn't translatable to a $match (cross-collection navigation traversal, a subquery, a GroupBy aggregate inside the predicate, etc.) passes this step but then fails when we actually build the filter. So a complex untranslatable predicate surfaces the same actionable error as an unsupported operator.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
ajcvickers: This path actually does go through MongoTransaction already. `database.BeginTransaction()` resolves our MongoTransactionManager, and that calls MongoTransaction.Start(...).

The only thing used directly is the session handle (CurrentTransaction.Session) passed into DeleteMany/UpdateMany, which is the same pattern MongoDatabaseWrapper uses for BulkWrite(session, …) in SaveChanges.

So there's no commit/rollback-logging or Guid gap here. That said, you're right that this shape doesn't belong in the query visitor; I'll move the whole two-phase orchestration into Storage (alongside MongoTransactionManager) as part of addressing your :548 comment.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
ajcvickers: Done. The bulk write/transaction/diagnostics execution logic now lives in MongoBulkOperationExecutor (Storage), driven by a MongoBulkPlan the visitor builds at compile time; the visitor keeps only the translation helpers and a thin VisitNonQuery/CreateBulkPlan hand-off. That moved ~500 lines out of the query-compiling visitor into Storage alongside the rest of the transaction/driver.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
ajcvickers: Fixed. VisitNonQuery no longer does the verbose two-block executor selection.

docs/agentplans/2026-06-09/execute-update-delete/execute-update-delete.design.md
```diff
@@ -0,0 +1,189 @@
+# ExecuteUpdate / ExecuteDelete for the MongoDB EF Core Provider — Design
+
+**Date:** 2026-06-09
+**Status:** Approved (design); pending implementation plan
+**JIRA:** TBD (`EF-NNN`)
```
damieng: Should put ticket number here if it's a historical ref.

docs/agentplans/2026-06-09/execute-update-delete/execute-update-delete.design.md
```diff
@@ -0,0 +1,189 @@
...
+(`RelationalQueryableExtensions`), which this provider deliberately does not reference, and the
+core base class `QueryableMethodTranslatingExpressionVisitor` has **no** `TranslateExecuteDelete`/
+`TranslateExecuteUpdate` virtuals in EF8. Therefore the feature is supportable only on **EF9 and
+EF10** and must be gated behind `#if !EF8`.
+
+The relational reference behavior (confirmed against the EF Core source):
+
+- `NonQueryResult` / `NonQueryResultAsync` in `RelationalShapedQueryCompilingExpressionVisitor`
+  issues exactly **one** SQL statement — no `BeginTransaction`, no `IDbContextTransaction`.
+  Atomicity of a standalone bulk op comes from the single statement being inherently atomic at
```
damieng, line 30: Do we wrap the query + bulk op in a transaction to ensure consistency between getting the list of ids and then bulk operating those ids in a single transaction?

docs/agentplans/2026-06-09/execute-update-delete/execute-update-delete.design.md
```diff
@@ -0,0 +1,189 @@
...
+  the database level.
+- The command **enlists in the context's `CurrentTransaction`** if one is already open.
```
damieng, line 32: We might need to call this out in the docs as MongoDB doesn't support nested transactions so this is a limitation if they're coming from some relational providers.

docs/agentplans/2026-06-09/execute-update-delete/execute-update-delete.design.md
```diff
@@ -0,0 +1,189 @@
...
+- `ExecuteUpdate(...)` / `ExecuteUpdateAsync(...)` on a single `DbSet`.
+- Source query scoped by `Where(...)` predicate(s) that the existing EF→driver-LINQ translator
+  can render.
+- TPH inheritance: the discriminator filter the query pipeline already injects **must** be
+  preserved in the bulk filter, so a bulk op on a derived type only affects documents of that
+  type.
+
+**Out of scope (v1) — must throw the canonical EF "could not be translated" message**
+
+- Any source-chain operator beyond filtering: `OrderBy`/`ThenBy`, `Skip`/`Take`, `Select`/
```
damieng, line 62: Is it a generic message or something specific to bulk operation limitations to point them in the right direction/avoid confusion when it's a query that normally works for returning results?

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -311,6 +381,302 @@ private static QueryingEnumerable<BsonDocument, TResult> ExecuteShapedQuery<TSou
...
+            .DeclaredMethods
+            .Single(m => m.Name == nameof(MongoBulkOperationExecutor.ExecuteAsync));
+
+    private static readonly MethodInfo RenderAggregateExpressionMethodInfo =
+        typeof(MongoShapedQueryCompilingExpressionVisitor)
+            .GetTypeInfo()
+            .DeclaredMethods
+            .Single(m => m.Name == nameof(RenderAggregateExpression));
+#endif
+
```
damieng, line 679: There's still quite a lot of code being added to here that feels like it could have a better home. Something that perhaps takes the MongoNonQueryExpression and builds all the necessary logic that's not related to shapedquerycompiling ?

src/MongoDB.EntityFrameworkCore/Storage/MongoBulkOperationExecutor.cs
```diff
@@ -0,0 +1,418 @@
...
+/// <c>deleteMany</c>/<c>updateMany</c> driver writes, orchestrates the two-phase transaction, materializes target
+/// ids, and emits the bulk diagnostics events. Translation of the filter/update/target query is deferred to the
+/// plan's delegates (owned by the query pipeline).
+/// </summary>
+internal static class MongoBulkOperationExecutor
+{
+    // Dispatch on the (strategy, kind) pair. A new MongoBulkStrategy (e.g. a future $merge path) adds an arm here;
+    // the explicit-arms-plus-throw shape routes any unhandled combination to a loud failure rather than silently
+    // falling through to single-command/delete.
+    public static int Execute(QueryContext queryContext, MongoBulkPlan plan)
```
damieng, line 47: Love this abstraction. Clearly says what it's going to deal with.

tests/MongoDB.EntityFrameworkCore.UnitTests/Query/Expressions/MongoNonQueryExpressionTests.cs
```diff
@@ -0,0 +1,122 @@
...
+public static class MongoNonQueryExpressionTests
+{
+    class Product
+    {
+        public ObjectId Id { get; set; }
+        public string Name { get; set; }
+        public decimal Price { get; set; }
+    }
+
+    class NonQueryDbContext : DbContext
```
damieng, line 34: Not sure we really get any value from unit tests any more. Perhaps we should tell Claude not to add new ones unless justified.

docs/agentplans/2026-06-09/execute-update-delete/execute-update-delete.design.md
ajcvickers: Done.

docs/agentplans/2026-06-09/execute-update-delete/execute-update-delete.design.md
ajcvickers: Yes. This design doc is the 2026-06-09 snapshot, where two-phase was still out-of-scope for v1; the shipped implementation now covers it.

docs/agentplans/2026-06-09/execute-update-delete/execute-update-delete.design.md
ajcvickers: Added docs. Note we never open a nested transaction: if the context already has a CurrentTransaction, the bulk op enlists in it. The only friction is AutoTransactionBehavior.Never + a two-phase shape with no open transaction, which throws an actionable error.

docs/agentplans/2026-06-09/execute-update-delete/execute-update-delete.design.md
ajcvickers: Both, depending on the failure. Out-of-scope operator chains (joins, projection, set ops, etc.) throw EF's canonical message, with details for the specific reason. The transaction-availability failures get bespoke messages that name the cause and the fix.

src/MongoDB.EntityFrameworkCore/Storage/MongoBulkOperationExecutor.cs
ajcvickers: Thanks! That extraction came out of your earlier review note about the logic not belonging in the query-compiling visitor.

tests/MongoDB.EntityFrameworkCore.UnitTests/Query/Expressions/MongoNonQueryExpressionTests.cs
ajcvickers: These cover MongoNonQueryExpression construction/equality/Update semantics, which the functional/spec suites don't exercise directly.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoShapedQueryCompilingExpressionVisitor.cs
ajcvickers: Keeping the translation logic here; the execution extraction is the part that mattered and it's done.

- Execution/transaction/diagnostics already moved to `MongoBulkOperationExecutor` (Storage) — `VisitNonQuery` now only builds the plan and emits the call.
- What remains is genuine query-translation (`BuildFilter`/`BuildUpdate`/`BuildIdDocumentQuery`) — Query AGENTS.md deliberately keeps that in Query, not Storage.
- EF Core's own `RelationalShapedQueryCompilingExpressionVisitor` handles non-query inline too, so the placement is idiomatic.
- A same-area `MongoNonQueryTranslator` split is possible but cosmetic, and would couple to the read path's `TranslateQuery` for no behavioral or boundary gain.

<details>
<summary>Full analysis</summary>

**What's actually in the visitor.** The file is 730 lines; the bulk/non-query code (all under `#if !EF8`) is ~330 of them, in three layers:

- **Entry/dispatch** (~75 lines) — `VisitExtension`→`VisitNonQuery`, `CreateBulkPlan<T>`, bulk `MethodInfo` fields. Compile-time: assemble a `MongoBulkPlan`, emit `Expression.Call` to the executor.
- **Validation/error mapping** (~28 lines) — `TranslateBulkOrThrow`, `EnsureBulkKeyOrThrow`.
- **Translation helpers** (~230 lines) — `BuildFilter`, `BuildUpdate`, `BuildIdDocumentQuery`, `SerializeConstant`, `RenderSelfReferencingValue`, `RenderAggregateExpression`, `EvaluateToConstant`, `CompileAndEvaluate`, `ParameterRebindingExpressionVisitor`.

No execution code remains here — writes, transactions, and diagnostics already live in `MongoBulkOperationExecutor` (Storage), the extraction from the earlier review round. `VisitNonQuery` only builds the plan and emits the call.

**Three facts that bear on the decision.**

1. The Query AGENTS.md explicitly documents the bulk seam as *deliberately* different from the read seam: *"The translation helpers (`BuildFilter`/`BuildUpdate`/`BuildIdDocumentQuery`) stay here in Query... Don't 'fix' the bulk seam to look like the read seam."* These use `MongoEFToLinqTranslatingExpressionVisitor`, `BsonSerializerFactory`, and render driver definitions — textbook Query-area work. Moving them to Storage would violate the documented boundary.
2. EF Core's own `RelationalShapedQueryCompilingExpressionVisitor` handles `NonQueryResult` inline in the shaped-query-compiling visitor — it does not farm it out. The current placement matches EF's reference architecture.
3. `BuildIdDocumentQuery` reuses the private static read-path `TranslateQuery<TSource>`. Any extraction has to expose or thread that through — real coupling cost, in a just-reviewed file.

**Verdict.** Parsed as a same-area `MongoNonQueryTranslator` (not a Storage move, which fact #1 rules out), the value is cosmetic: it would drop the visitor to ~400 focused lines, but trade "long file" for "two coupled files," re-churn a freshly-reviewed area, and require a full EF8/9/10 re-test for no behavioral or boundary gain. The high-leverage refactor (execution → executor) is already done; what remains is correctly-placed translation logic.

</details>

ajcvickers:
@damieng Are you still waiting for changes on this? (I just rebased on the search index work.)

damieng (approved):

PR #318, opened 2026-06-18
EF-321: Design and plan for geospatial query support (v1)

ajcvickers:
- **Storage:** entities use the driver's GeoJSON types directly; serialization is free via the driver's existing serializers.
- **Indexing:** `Is2dSphereIndex()` configures a `2dsphere` index that flows through the normal non-vector `CreateIndexDocument()` path.
- **API:** three predicates — `GeoWithin`, `GeoIntersects`, `GeoWithinCircle` — exposed as `EF.Functions` markers.
- **Translation:** `MongoEFToLinqTranslatingExpressionVisitor` intercepts the enclosing `Where` and appends a raw geo `$match` via `MongoQueryable.AppendStage` (the `ProcessVectorSearch` mechanism), since the driver has no LINQ geo translation.
- **Constraint:** queries run as aggregations, so `$near` is out — proximity is `$centerSphere` containment, with no distance sort.
- **Extensibility:** filter-BSON construction is isolated in a type-agnostic `MongoGeoTranslator`, the seam for a later NetTopologySuite layer.

### DIFF
--- a/docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
+++ b/docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
@@ -1,0 +1,249 @@
+# Geospatial query support (driver GeoJSON types) — Design
+
+**Date:** 2026-06-18
+**Status:** Approved (design); implementation plan written
+**JIRA:** EF-321
+
+## Summary
+
+Add MongoDB geospatial support to the EF Core provider in a tractable first release:
+
+1. Allow MongoDB C# driver GeoJSON types (`GeoJsonPoint<T>`, `GeoJsonPolygon<T>`, …) on
+   entity properties, persisted via the driver's existing GeoJSON serializers.
+2. Create `2dsphere` indexes over those properties through a fluent model-building API and
+   `EnsureCreated`.
+3. Translate three geospatial predicates — exposed as `EF.Functions` methods — into the
+   corresponding MQL geo `$match` operators.
+
+The design is deliberately staged so a future NetTopologySuite (NTS) layer can be added on
+top of a shared, type-agnostic translation core without rework.
+
+## Background and key constraints
+
+Two findings shaped this design and override the assumptions in the initial exploration:
+
+1. **The driver has no LINQ geo translation.** The MongoDB C# driver (v3.9.0) ships GeoJSON
+   *serializers* (`GeoJsonPointSerializer<T>`, etc.), so persisting GeoJSON values is free.
+   But its geo *query* operators (`GeoWithin`, `GeoIntersects`, `NearSphere`, `GeoNear`) are
+   exposed only through the filter-definition builder (`Builders<T>.Filter.GeoWithin(...)`)
+   and the aggregate-fluent `GeoNear` stage — there is **no `IMongoQueryable` geo extension**.
+   A user cannot write `.Where(x => x.Location.GeoWithin(poly))` in driver LINQ. Therefore the
+   provider must translate geo predicates **explicitly** (emit raw geo `$match` BSON), the same
+   way the VectorSearch feature is handled — not by relying on LINQ passthrough.
+
+2. **The provider executes queries as aggregation pipelines** (`IMongoCollection.Aggregate`,
+   confirmed in `Query/SerializerOverrideCollection.cs`). MongoDB forbids `$near` / `$nearSphere`
+   inside an aggregation `$match`. Proximity inside a pipeline must use either `$geoWithin` with
+   `$centerSphere` (radius containment, no sorting) or the `$geoNear` stage (must be first stage,
+   produces a computed distance). Consequently, v1 proximity is **radius containment**, not
+   distance-sorted "near". The distance-sorted / computed-distance behavior is deferred to a
+   later `$geoNear` phase.
+
+## Type-system decision
+
+v1 exposes the **driver's GeoJSON types** directly on entities. NetTopologySuite (NTS) support —
+the EF-ecosystem standard that would satisfy EF Core's NTS-based spatial conformance suite — is
+deferred to a later phase. Rationale:
+
+- Driver GeoJSON serialization is free; NTS would require a bidirectional NTS↔GeoJSON conversion
+  serializer plus coordinate-system / SRID handling (NTS is planar `(X, Y)` + SRID; GeoJSON is
+  spherical `[lng, lat]`).
+- With driver types we control a small, invented predicate vocabulary (`EF.Functions.*`); NTS
+  would require mapping NTS's larger external method surface for conformance.
+
+The required explicit-translation core (finding 1 above) is type-agnostic and becomes the seam
+the NTS phase plugs into (see "NTS seam").
+
+## Scope
+
+### In scope (v1)
+
+| `EF.Functions` method | MQL emitted | Notes |
+|---|---|---|
+| `GeoWithin(field, area)` | `{ field: { $geoWithin: { $geometry: <area> } } }` | `area` is any GeoJSON polygon / multipolygon |
+| `GeoIntersects(field, geometry)` | `{ field: { $geoIntersects: { $geometry: <geometry> } } }` | geometries overlap |
+| `GeoWithinCircle(field, center, radiusMeters)` | `{ field: { $geoWithin: { $centerSphere: [ [lng,lat], radiusMeters/6378100 ] } } }` | spherical radius containment |
+
+Plus:
+- Persisting driver GeoJSON types on entity properties.
+- `2dsphere` index creation via a fluent builder API and `EnsureCreated`.
+
+### Out of scope (deferred)
+
+- `$geoNear` aggregation stage.
+- Distance projection / distance-based ordering (a true `$near`).
+- `minDistance` (inner-radius exclusion) — v1 takes a single `radiusMeters`.
+- Legacy `2d` indexes.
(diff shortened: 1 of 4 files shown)


docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
```diff
@@ -0,0 +1,249 @@
...
+Users place driver GeoJSON types directly on entities:
+
+```csharp
+public GeoJsonPoint<GeoJson2DGeographicCoordinates> Location { get; set; }
+public GeoJsonPolygon<GeoJson2DGeographicCoordinates> Area { get; set; }
+```
+
+### Predicates
+
+A new `MongoDbFunctionsExtensions` static class extends `DbFunctions` (so the methods hang off
```
damieng, line 99: Why would these be on a new MongoDbFunctionsExtensions vs using our own Mql class where we normally put extensions?  The latter would also give us a shared entry point for this work to be done later on the C# Driver LINQ side.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
```diff
@@ -0,0 +1,249 @@
...
+Each method body throws `InvalidOperationException` — the standard `EF.Functions`
+"translation-only marker" pattern. The shared `TCoordinates` generic prevents mixing coordinate
+systems at compile time.
+
+### Index
+
+A fluent extension in `MongoIndexBuilderExtensions`:
+
+```csharp
+modelBuilder.Entity<Place>().HasIndex(e => e.Location).Is2dSphereIndex();
```
damieng, line 125: Are we going to create the index? What does this fluent API trigger behind the scenes?

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
```diff
@@ -0,0 +1,249 @@
...
+- Tests (see "Testing").
+
+### Edited files
+
+- `Metadata/MongoAnnotationNames.cs` — add `GeospatialIndexType` constant
+  (`"Mongo:GeospatialIndexType"`).
+- `Extensions/MongoIndexBuilderExtensions.cs` — add `Is2dSphereIndex()` (generic and non-generic,
+  matching the existing `IsVectorIndex` pair).
+- `Metadata/InternalIndexExtensions.cs` — `CreateIndexDocument()` detects the annotation and emits
+  `{ "<element-path>": "2dsphere" }`.
```
damieng, line 155: Is this the part picking up the index annotation and then providing they're running CreateDatabase/CreateIndexes is being emitted?

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+The geometry argument is a captured value, not a column: it is evaluated to a concrete GeoJSON
+object at translation time and serialized, exactly as other constant/parameter arguments are
+handled today. `radiusMeters` is read as a value and converted to radians (`radiusMeters / 6378100`,
+using the equatorial Earth radius in meters that MongoDB documents for `$centerSphere`).
+
+## Index creation (`2dsphere`)
+
+`Is2dSphereIndex()` sets `Mongo:GeospatialIndexType = "2dsphere"` on the `IMutableIndex`. During
```
damieng, line 187: Ah, questions answered.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
```diff
@@ -0,0 +1,1315 @@
...
+            options.Unique ??= index.IsUnique;
+        }
+
+        return new CreateIndexModel<BsonDocument>(doc, options);
+    }
+```
+
+- [ ] **Step 4: Update `MakeIndexName` for geospatial indexes**
+
+In `InternalIndexExtensions.cs`, in `MakeIndexName(this IReadOnlyIndex index)` (lines 28-48), after the existing vector-index early return (line 31-34), add:
```
damieng, line 331: Is this naming based on any existing pattern?  Driver-generated index names are supposed to be consistent across drivers.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
```diff
@@ -0,0 +1,1315 @@
...
+Expected: FAIL — EF rejects the `GeoJsonPoint<>` property (no type mapping) at model build, or no serializer is found.
+
+- [ ] **Step 3: Recognize GeoJSON types in the type-mapping source**
+
+In `MongoTypeMappingSource.cs`, add a static helper and extend `FindPrimitiveMapping`. Add near the top of the class (after the `SupportedDictionaryInterfaces` array, ~line 45):
+
+```csharp
+    private static bool IsGeoJsonType(Type type)
+    {
+        for (var t = type; t is not null; t = t.BaseType!)
```
damieng: I think we have a routine already to see if something in the hierarchy matches that we use for List<T> etc.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
```diff
@@ -0,0 +1,1315 @@
+# Geospatial query support (driver GeoJSON types) — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Let users persist MongoDB driver GeoJSON types on entities, create `2dsphere` indexes over them, and filter with three `EF.Functions` geospatial predicates that translate to MQL geo `$match` stages.
```
damieng, line 5: The goal is to allow persistence but I mostly saw query, deserialization and index work.  Is there an actual serialize+save path described somewhere I missed?

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
damieng: I left this and the previous ones in to show my thought process going through the PR in case we want to optimize the design docs to be more easily answered as-read.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
ajcvickers: If we want to do it this way, then we would have to do the driver work first and then wait for it to be shipped. So this probably means we can't really do this feature without driver work first. However, for the sake of this experiment which will likely not even ship, let's assume that, for once, we are not blocked on the driver.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
ajcvickers: Index names in general are not consistent for different index types even within a single driver. That being said, I can try to find out if there is a de facto or de jury convention for this anywhere.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
ajcvickers: Good feedback.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
ajcvickers: GeoJSON properties serialize through the same `BsonSerializerFactory.CreateTypeSerializer` arm the plan adds (Task 4, Step 4). On save, `MongoUpdate.WriteProperty` resolves the serializer via GetPropertySerializationInfo and calls `Serializer.Serialize(...)` to write BSON . So one serializer registration enables both insert/update writes and reads

I'l update the design doc to include this explicitly.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
damieng: The basic indexes are in terms of field order, etc. but yeah it doesn't cover all cases.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.design.md
ajcvickers: Added a forward-reference from the Is2dSphereIndex() API to the Index-creation section so the "does this create the index?" question is answered where it's raised.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
ajcvickers: Switched to the dotted element-path key + _2dsphere (e.g. address.location_2dsphere), which matches the name MongoDB auto-generates for an unnamed 2dsphere index, so EF- and server-created names agree.

docs/agentplans/2026-06-18/geospatial-queries/geospatial-queries.implementation.md
ajcvickers: Confirmed and reused `TryGetItemType(typeof(GeoJsonGeometry<>))`

damieng (approved):

ajcvickers:
Closing this for now in favor of doing the LINQ work first.

PR #322, opened 2026-06-29
EF-323: Native query-translation foundation (native filter / sort / paging)

ajcvickers:
Sub-project 1 of the ground-up native LINQ query rebuild (EF-322): the provider now translates queries to MongoDB aggregation pipelines itself for the supported slice, instead of always delegating to the C# driver's LINQ provider.

- Translates single-collection **filter / sort / paging** over whole-entity results to native MongoDB aggregation pipelines (`$match`/`$sort`/`$skip`/`$limit`, executed via `Aggregate`); everything outside that slice transparently falls back to the driver's LINQ provider.
- New compile-time gate `MongoQueryMode` — `Native` (default), `DriverLinq` (pre-rebuild behaviour), `NativeOnly` (diagnostic: native-or-throw) — selected via `UseQueryMode(...)`.
- ~50–58% lower allocation and ~18–51% faster on the native slice (quiet M4 Max, N=10k), matching the reference spike's native numbers.
- Purely additive public API (`MongoQueryMode`, `UseQueryMode`); query results are identical to before; no breaking changes.
- Zero regressions — full Unit / Functional / Specification suites green on EF8, EF9 and EF10.
- Unsupported shapes are detected and fall back, never mis-translated — proven by native-vs-driver-LINQ parity tests across the mis-routing surface.
- Reference `Include`, predicate breadth, `$project` pushdown, scalar cardinality and the operator long tail are deferred to later sub-projects (they fall back today).

**Pipeline (built the EF / relational way).** The QMTEV populates structured slots on `MongoQueryExpression` (predicate / orderings / offset / limit) instead of returning `null`; `MongoSelectLowerer` lowers them to a typed `MongoPipelineStage[]` IR; `MongoQueryLanguageRenderer` renders the `$match` body (with the driver-parity AND/OR field-merge) and `MongoPipelineFactory` walks the stages into a `BsonDocument[]` template.

**"B2" translation timing.** Lowering + rendering happen **once at compile time** into a cached pipeline *template* whose `MongoParameterExpression` placeholders are bound per execution — so per-query work is a clone-and-substitute, not a re-walk of the LINQ chain. A consequence: the native-vs-driver and streaming-vs-DOM decisions are made deterministically at **compile time**, so exactly one shaper is compiled per query (no runtime dual-shaper dispatch).

**Conservative gate.** The driver-LINQ path is retained unchanged as the gated fallback, and the QMTEV always also captures the raw method chain so the fallback stays intact. Any operator the native path does not lower into a slot is marked non-representable and falls back — correctness-safe by construction (a missed optimisation, never a wrong result). Under `NativeOnly` such a query throws instead (a coverage/diagnostic instrument).

**Materialization.** Whole-entity results for streaming-eligible entities materialize via a forward-only reader over `RawBsonDocument` (no intermediate DOM); other shapes use the existing `BsonDocument` shaper. The streaming materializer is reproduced from the reference spike (which is reference-only; nothing else was ported).

**Scope / deferral.** Native single-level reference `Include` was in the original SP1 target but is **deferred to the Collection Includes sub-project**: EF nav-expands it into a `LeftJoin` the gate treats as non-native, so it falls back to driver-LINQ (correct results; throws under `NativeOnly`). The `$lookup`/`$unwind` lowering + streaming-Include machinery is built but dormant, ready for that sub-project to activate.

**Notable as-built decisions.** The slot model was added *in place* on `MongoQueryExpression` (rather than introducing a separate `MongoSelectExpression`) to avoid churning the QMTEV/shaper/factory plumbing during a zero-regression foundation; the native carriers on `MongoExecutableQuery` (`NativePipeline`/`Session`/`Streaming`) are `internal`; and `MongoQueryMode` participates in `GetServiceProviderHashCode`/`ShouldUseSameServiceProvider` so contexts differing only by mode don't share a service provider or compiled-query cache.

**Correctness.** Validated against driver-LINQ as the oracle. Four divergences found during review were fixed, each with a committed regression test: (1) paging was silently *not* going native; (2) `Where`/`OrderBy` after `Skip`/`Take` reordered stages and returned wrong rows; (3) non-positive `Take` now throws `ArgumentOutOfRangeException` (matching EF/driver) rather than emitting an invalid `{$limit:0}`; (4) two streaming-materializer nullability cases — a *missing* required scalar now throws `InvalidOperationException` (matching the driver entity binding) and an explicit BSON `null` on a non-nullable property now materializes `default(T)` (matching the driver) instead of throwing.

<details>
<summary>Full detail — components, files, scope, testing, performance</summary>

- **`Expressions/MongoExpression`** hierarchy (`MongoField`/`Constant`/`Parameter`/`Binary`/`Unary` + `MongoOrdering`) — a dialect-agnostic `SqlExpression`-analog node set (custom `System.Linq.Expressions.Expression` subclasses; the visitor machinery is present for future passes but not yet exercised at parity).
- **`Expressions/MongoQueryExpression.NativeSlots.cs`** — structured slots (`Predicate`/`Orderings`/`Offset`/`Limit`/`IsNativeRepresentable`/`Lookups`) added to the existing query root; retains `CapturedExpression` + `ProjectionMapping` for the fallback.
- **`Visitors/MongoQueryableMethodTranslatingExpressionVisitor`** — `PopulateNativeSlots` fills the slots for `Where`/`OrderBy(Descending)`/`ThenBy(Descending)`/`Skip`/`Take`; everything else (a non-slot operator, an un-translatable predicate, a non-canonical order, a projecting `Select`, `OfType`, composite-PK key access, …) sets `IsNativeRepresentable = false`. Always still captures the raw chain.
- **`NativeTranslation/MongoExpressionTranslator`** — EF predicate / sort-key body → `MongoExpression` (carries the spike's element-name resolution + value coercion; query parameters become placeholders, not baked constants); acceptance set matches the spike's (nullable-equality, numeric member casts, method calls, composite-PK, etc. fall back).
- **`NativeTranslation/MongoSelectLowerer`** + **`Stages/`** — slots → typed `MongoPipelineStage[]` in canonical order `$match → $sort → $skip → $limit → $lookup/$unwind`; BSON-free; owns the single-level-reference lookup eligibility (`LookupExpression.IsStreamableReference`).
- **`NativeTranslation/MongoQueryLanguageRenderer`** + **`PlaceholderTable`** — `MongoExpression` → `$match`-dialect BSON; records parameter placeholders. `MongoExprRenderer`/dialect selection is **deferred to SP2** (no `$expr` renderer at the foundation).
- **`NativeTranslation/BsonValueSerializer`** — shared coerce + serialize-through-`IBsonSerializer` helper used by both the renderer (constants) and the factory (parameters); property-less primitives (Skip/Take counts) serialize via `BsonValue.Create`.
- **`NativeTranslation/MongoPipelineFactory`** — `Create(stages, renderer)` walks the IR into a `BsonDocument[]` template + placeholder table; `Build(parameterValues)` deep-clones and substitutes per execution, validating `$limit > 0` / `$skip ≥ 0`.
- **`NativeTranslation/{MongoStreamingEntityMaterializerRewriter, BsonRowReader, StreamingEligibility}`** — forward-only `RawBsonDocument` → POCO materialization (reproduced from the spike; missing/null-required handling fixed to match driver-LINQ).
- **`NativeTranslation/NativeQueryParameter`** — single EF8/EF9-vs-EF10 query-parameter detection helper (used by the translator + the QMTEV).

- **`Visitors/MongoShapedQueryCompilingExpressionVisitor`** — the compile-time gate: under `Native`, if `IsNativeRepresentable` and lower/render succeed it builds the `MongoPipelineFactory` and compiles the streaming or DOM shaper; otherwise (or under `DriverLinq`) it compiles the driver-LINQ fallback; under `NativeOnly` a non-representable query throws.
- **`MongoExecutableQuery`** gains `internal` `NativePipeline`/`Session`/`Streaming`; **`Storage/MongoClientWrapper.Execute`** runs `collection.Aggregate(pipeline)` (raw or DOM, session-aware) when a native pipeline is present, else the driver-LINQ path; MQL is logged through the existing redaction-gated `ExecutedMqlQuery` overload.
- **`Infrastructure/MongoQueryMode`** + **`MongoDbContextOptionsBuilder.UseQueryMode`** + `MongoOptionsExtension.QueryMode`/`WithQueryMode` + `MongoQueryCompilationContext.QueryMode` — the user-facing option, folded into the options hash/equality/debug-info.

- **In (native):** `Where` (the basic predicate set `== != < <= > >= && ||`, bare/negated bool), `OrderBy`/`ThenBy`(`Descending`), canonical `Skip`/`Take`, whole-entity results.
- **Falls back to driver-LINQ (deferred to later sub-projects):** reference `Include`; predicate breadth (string methods, `Contains`/`IN`, computed, date, member casts, nullable-equality); `$project` pushdown; scalar cardinality (`Count`/`First`/`Any`/aggregates); `GroupBy`/`SelectMany`/set-ops/`Distinct`/`OfType`/`VectorSearch`; non-canonical paging; composite-PK key predicates/sort.

- New unit tests for every translation component (`tests/.../UnitTests/Query/NativeTranslation/`).
- New functional tests: `QueryModeGateTests`, `QueryModeGateIncludeTests`, `QueryModeOptionTests`, `NativePipelineExecutionTests`, `NativeGateRoutingTests` (value-converter / owned-nav / TPH native-vs-driver parity), `NativeCombinationQueryTests` (multi-key ordered + parameterized paging under `NativeOnly`), `NativeTransactionAndCancellationTests` (native-in-transaction + async cancellation mid-stream), `NativeMaterializerNullabilityTests` (missing/null required parity).
- `MongoTestStore` gains a **test-only** `MONGODB_EF_NATIVE_ONLY=1` env gate that flips the spec suite to `NativeOnly` — a native-coverage instrument (default-off; zero effect on normal runs).
- Verified on EF8, EF9 and EF10 (full Unit/Functional/Specification suites, 0 failures).

| Shape | Time | Allocation |
|---|---|---|
| Where→ToList | 8.27 vs 15.15 ms (−45%) | 9.6 vs 22.7 MB (−58%) | | Whole-entity no-track | 15.81 vs 32.37 ms (−51%) | 19.1 vs 45.3 MB (−58%) | | Whole-entity tracked | 25.22 vs 42.68 ms (−41%) | 25.2 vs 51.4 MB (−51%) | | OrderByTake | 2.61 vs 3.20 ms (−18%) | 0.24 vs 0.51 MB (−53%) | | Reference Include | ~parity (falls back — deferred) | ~same |

Native matches the reference spike's recorded native numbers on the supported slice (the SP1 "native ≥ spike" bar). A third `EF-Native` config was added to the benchmark harness; `results/perf-baseline.md` records the authoritative numbers.

`Query/AGENTS.md` rewritten for the dual-path architecture; SP1 design/overview docs record the scope and the as-built deferrals. (The SP1 design/overview/plan docs also appear in the docs PR; dedupe during rebase.)

</details>

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Infrastructure/MongoOptionsExtension.cs
+++ b/src/MongoDB.EntityFrameworkCore/Infrastructure/MongoOptionsExtension.cs
@@ -21,6 +21,7 @@ using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.Extensions.DependencyInjection;
 using MongoDB.Driver;
 using MongoDB.EntityFrameworkCore;
+using MongoDB.EntityFrameworkCore.Infrastructure;
 
 // ReSharper disable once CheckNamespace (extensions should be in the EF namespace for discovery)
 namespace Microsoft.EntityFrameworkCore;
@@ -53,6 +54,7 @@ public class MongoOptionsExtension : IDbContextOptionsExtension
         CryptProviderPath = copyFrom.CryptProviderPath;
         KeyVaultNamespace = copyFrom.KeyVaultNamespace;
         KmsProviders = copyFrom.KmsProviders;
+        QueryMode = copyFrom.QueryMode;
         _loggableConnectionString = SanitizeConnectionStringForLogging(ConnectionString);
     }
 
@@ -226,6 +228,25 @@ public class MongoOptionsExtension : IDbContextOptionsExtension
     }
 
     /// <summary>
+    /// Obtains the current <see cref="MongoQueryMode"/> that controls how LINQ queries are translated.
+    /// </summary>
+    public MongoQueryMode QueryMode { get; private set; } = MongoQueryMode.Native;
+
+    /// <summary>
+    /// Specifies the <see cref="MongoQueryMode"/> to use when translating LINQ queries.
+    /// </summary>
+    /// <param name="queryMode">The <see cref="MongoQueryMode"/> to use.</param>
+    /// <returns>The <see cref="MongoOptionsExtension"/> to continue chaining configuration.</returns>
+    public virtual MongoOptionsExtension WithQueryMode(MongoQueryMode queryMode)
+    {
+        Check.IsDefinedOrNull<MongoQueryMode>(queryMode);
+
+        var clone = Clone();
+        clone.QueryMode = queryMode;
+        return clone;
+    }
+
+    /// <summary>
     /// Clones the current <see cref="MongoOptionsExtension"/>.
     /// </summary>
     /// <returns>A new clone.</returns>
@@ -283,7 +304,7 @@ public class MongoOptionsExtension : IDbContextOptionsExtension
         /// <inheritdoc/>
         public override int GetServiceProviderHashCode()
         {
-            _serviceProviderHash ??= HashCode.Combine(Extension.ConnectionString, Extension.DatabaseName);
+            _serviceProviderHash ??= HashCode.Combine(Extension.ConnectionString, Extension.DatabaseName, Extension.QueryMode);
             return _serviceProviderHash.Value;
         }
 
@@ -294,7 +315,8 @@ public class MongoOptionsExtension : IDbContextOptionsExtension
                && Extension.MongoClient == otherInfo.Extension.MongoClient
                && (Extension.ClientSettings == otherInfo.Extension.ClientSettings ||
                    Extension.ClientSettings != null && Extension.ClientSettings.Equals(otherInfo.Extension.ClientSettings))
-               && Extension.DatabaseName == otherInfo.Extension.DatabaseName;
+               && Extension.DatabaseName == otherInfo.Extension.DatabaseName
+               && Extension.QueryMode == otherInfo.Extension.QueryMode;
 
         /// <inheritdoc/>
         public override void PopulateDebugInfo(IDictionary<string, string> debugInfo)
@@ -303,6 +325,7 @@ public class MongoOptionsExtension : IDbContextOptionsExtension
             AddDebugInfo(debugInfo, nameof(MongoClient), Extension.MongoClient);
             AddDebugInfo(debugInfo, nameof(ClientSettings), Extension.ClientSettings);
             AddDebugInfo(debugInfo, nameof(DatabaseName), Extension.DatabaseName);
+            AddDebugInfo(debugInfo, nameof(QueryMode), Extension.QueryMode);
         }
 
         /// <inheritdoc/>
@@ -335,6 +358,12 @@ public class MongoOptionsExtension : IDbContextOptionsExtension
             }
 
             builder.Append("DatabaseName=").Append(Extension.DatabaseName).Append(' ');
+
+            if (Extension.QueryMode != MongoQueryMode.Native)
+            {
+                builder.Append("QueryMode=").Append(Extension.QueryMode).Append(' ');
+            }
+
             return builder.ToString();
         }
     }
--- a/src/MongoDB.EntityFrameworkCore/Infrastructure/MongoQueryMode.cs
+++ b/src/MongoDB.EntityFrameworkCore/Infrastructure/MongoQueryMode.cs
@@ -1,0 +1,29 @@
+/* Copyright 2023-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB.EntityFrameworkCore.Infrastructure;
+
+/// <summary>Selects how the provider translates LINQ queries to MongoDB.</summary>
+public enum MongoQueryMode
+{
+    /// <summary>Native translation when representable; driver-LINQ fallback otherwise. (Default.)</summary>
+    Native,
+
+    /// <summary>Always use the driver's LINQ provider (the pre-rebuild behavior).</summary>
+    DriverLinq,
+
+    /// <summary>Native translation only; throw at compile time on an un-representable query (diagnostic).</summary>
+    NativeOnly
+}
--- a/src/MongoDB.EntityFrameworkCore/Query/Expressions/LookupExpression.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Expressions/LookupExpression.cs
@@ -67,7 +67,10 @@ internal sealed class LookupExpression
     /// <param name="navigation">The navigation the lookup supports.</param>
     /// <returns>The <c>_lookup_&lt;NavigationName&gt;</c> field name.</returns>
     public static string GetLookupAlias(IReadOnlyNavigationBase navigation)
-        => $"_lookup_{navigation.Name}";
+        => $"{LookupAliasPrefix}{navigation.Name}";
+
+    /// <summary>The prefix of the synthetic <c>$lookup</c> alias field (see <see cref="GetLookupAlias"/>).</summary>
+    public const string LookupAliasPrefix = "_lookup_";
 
     /// <summary>The navigation this lookup supports.</summary>
     public INavigation Navigation { get; }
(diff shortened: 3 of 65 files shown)


src/MongoDB.EntityFrameworkCore/Infrastructure/MongoQueryMode.cs
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB.EntityFrameworkCore.Infrastructure;
+
+/// <summary>Selects how the provider translates LINQ queries to MongoDB.</summary>
+public enum MongoQueryMode
```
ajcvickers, line 19: MongoQueryProvider

src/MongoDB.EntityFrameworkCore/Infrastructure/MongoOptionsExtension.cs
```diff
@@ -225,6 +227,25 @@ public virtual MongoOptionsExtension WithQueryableEncryptionSchemaMode(Queryable
...
+    /// Obtains the current <see cref="MongoQueryMode"/> that controls how LINQ queries are translated.
+    /// </summary>
+    public MongoQueryMode QueryMode { get; private set; } = MongoQueryMode.Native;
+
+    /// <summary>
+    /// Specifies the <see cref="MongoQueryMode"/> to use when translating LINQ queries.
+    /// </summary>
+    /// <param name="queryMode">The <see cref="MongoQueryMode"/> to use.</param>
+    /// <returns>The <see cref="MongoOptionsExtension"/> to continue chaining configuration.</returns>
+    public virtual MongoOptionsExtension WithQueryMode(MongoQueryMode queryMode)
```
ajcvickers, line 240: WithQueryProvider

src/MongoDB.EntityFrameworkCore/Query/Expressions/LookupExpression.cs
```diff
@@ -118,6 +121,13 @@ private static string GetFieldPath(IReadOnlyProperty property)
     /// <summary>Whether this lookup is for a single reference (not a collection).</summary>
     public bool IsReference => !Navigation.IsCollection;
 
+    /// <summary>
+    /// A single-level reference Include the native pipeline can emit and the streaming reader can read back:
+    /// a reference nav, no filtered-Include pipeline stages, not a transitive <c>_lookup_</c> local field.
+    /// </summary>
+    public bool IsStreamableReference
```
ajcvickers, line 128: Is using the well-known prefix a good idea here?

src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoExpression.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <inheritdoc />
+    protected override Expression VisitChildren(ExpressionVisitor visitor)
+        => this;
+}
+
+/// <summary>
+/// Binary comparison and logical operators for <see cref="MongoBinaryExpression"/>.
+/// </summary>
+internal enum MongoBinaryOperator
```
ajcvickers, line 46: Own file. And below.

src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoParameterExpression.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>The parameter name.</summary>
+    public string Name { get; }
+
+    /// <summary>
+    /// Optional property metadata used by the renderer to select the correct serializer.
+    /// </summary>
+    public IProperty? ForSerialization { get; }
+
+    /// <inheritdoc />
+    public override Type Type
```
ajcvickers, line 53: Why is the type object if the IProperty is not present?

src/MongoDB.EntityFrameworkCore/Query/Expressions/LookupExpression.cs
ajcvickers: This can be hardened. Recorded on EF-317.

src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoParameterExpression.cs
ajcvickers: ### When `ForSerialization` is null                                                                                                                                                                                                                                                                              
  Only when the parameter has no owning mapped field:                                                                                                                                                                                                                                                              
  1. **Paging counts** — `Skip(n)`/`Take(m)` captured vars (`NativeSlotPopulator.TranslateCountExpression`).                                                                                                                                                                                                       
  2. **Bare operands in an `$expr` arithmetic / field-to-field subtree** — e.g. `p` in `x.A + p` (`MongoExpressionTranslator.TranslateValue(..., forSerialization: null)`).                                                                                                                                        
                                                                                                                                                                                                                                                                                                                   
  Non-null (common case): parameter compared to a mapped field (`x.Age == v`), `$in`/`$nin` values, `StartsWith`/regex terms — the field's `IProperty` is threaded for serializer selection.                                                                                                                       
                                                                                                                                                                                                                                                                                                                   
  ### What `Type == object` means functionally                                                                                                                                                                                                                                                                     
  - **`.Type` is inert for parameters** — nothing in the native path reads `MongoParameterExpression.Type`; it just satisfies the `MongoExpression` base contract. `object` = "type-erased/unknown."                                                                                                               
  - **The real effect rides on `ForSerialization == null`:** renderer emits a serializer-less placeholder (`CreatePlaceholder(name, serializer: null)`), and at execution `MongoPipelineFactory.Build` substitutes via `BsonValue.Create(rawValue)` — bypassing any property serializer (value converter,          
  `BsonRepresentation`, element serializer). Correct because these values have no owning property; it's also why comparison/`$in` paths must thread the `IProperty`.                                                                                                                                               
                                                                                                                                                                                                                                                                                                                   
  **Latent edge (not a current bug):** the serializer-less path is only sound for property-less primitives. If `$expr` operand support ever broadened to value-converted properties via the `null` branch, the converter would be silently skipped — the numeric-only guard prevents that today.

PR #323, opened 2026-07-01
EF-149: Support GroupBy aggregations

damieng:
Translate single-collection GroupBy to MongoDB $group via the driver's LINQ provider. TranslateGroupBy now returns a GroupByShaperExpression which TranslateSelect / the projection binder collapse into a scalar or anonymous projection of the key and aggregates; the captured chain the driver receives already renders to $group/$project.

Supports scalar/composite/anonymous keys, element and result selectors, the standard aggregates (Count/LongCount/Sum/Min/Max/Average), and post-group Where/OrderBy/Select/Distinct.

GroupBy composed with Join/GroupJoin/subqueries, grouping by a cross-collection navigation, and per-group entity (IGrouping) materialization remain unsupported and are rejected cleanly. Joining over a grouped query previously returned silent wrong results; it now fails translation via the IsGroupByQuery guard.

Flips the NorthwindGroupByQueryMongoTest conformance suite from asserting translation failure to asserting the generated $group pipeline (green on EF8/EF9/EF10), and updates the affected SetOperations/KeylessEntities/AggregateOperators/Miscellaneous overrides. EF-149 failing-spec count 246 -> 109.


docs/failing-spec-tests.md
ajcvickers: Claude says EF-216 count stale. The PR drops 7 EF-216 tags but the EF-216 row count stays at 265.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -636,6 +652,50 @@ private void ExitProjectionMember()
...
+    private static bool IsGroupByAggregate(MethodCallExpression call)
+    {
+        if (call.Arguments.Count == 0)
+        {
+            return false;
+        }
+
+        switch (call.Method.Name)
+        {
+            case nameof(Enumerable.Count):
```
ajcvickers: Better in general to match on MethodInfo rather than name.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -636,6 +652,50 @@ private void ExitProjectionMember()
     /// being fully visited. This covers <c>EF.Property</c> (for non-navigation properties) and
     /// <c>Mql.Field</c> calls.
     /// </summary>
```
ajcvickers: Duplicate doc comment.

ajcvickers:
Make sure that Agents.md, readme, etc. are updated to record that these GroupBy shapes are now supported.

docs/failing-spec-tests.md
damieng: Now fixed.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
damieng: Done.

src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
damieng: Corrected.

PR #324, opened 2026-07-07
Native query working branch

ajcvickers:
Native MQL translation for the MongoDB EF Core provider — the working branch for epic [EF-322](https://jira.mongodb.org/browse/EF-322).

**Still a draft, and deliberately so.** The remaining work is now *parked* rather than in flight: every known gap has a ticket, and every loose end in the code points at one. What is here is complete and green; what is not here is enumerated below.

## What this changes

The provider gains a **second query path**. For the supported slice it builds the aggregation pipeline itself — `MongoExpression` IR → `MongoSelectLowerer` → `MongoQueryLanguageRenderer` / `MongoAggregationExpressionRenderer` → `BsonDocument[]` → `IMongoCollection.Aggregate` — instead of handing a LINQ expression to the driver's LINQ v3 provider. Anything outside that slice falls back to the driver automatically.

Native is the **default**. Per the provider's versioning rubric that is not in itself a breaking change: results are unchanged, unsupported shapes fall back, and `UseQueryMode(MongoQueryMode.DriverLinq)` restores the previous path. Three places where something *observable* did change carry `BREAKING-CHANGES.md` entries — see below.

## Size and shape

| | |
|---|---|
| Base | `58e05a0e` (EF-317, driver 3.10.0) — **0 behind** |
| Commits | 103 |
| `src/` | 94 files (71 new), +19,321 / −242 |
| `tests/` | 105 files, +47,082 / −860 |
| `docs/` | 114 files, +59,322 / −93 |

`docs/superpowers` carries the spikes and implementation plans, following the existing `docs/agentplans` convention on `main`.

## Where it stands, measured

Two axes, both on EF10, both re-measured at the tip:

- **Default `Native`** — **4593 passed / 0 failed / 17 skipped**. This is the axis that matters for "did anything regress": the fallback is live, so every supported shape must still return the same answers.
- **`MONGODB_EF_NATIVE_ONLY=1`** — **2507 passed / 2086 failed / 17 skipped**. This is the coverage report: the 2086 are the specification cases that still *need* driver-LINQ. It is a deliberate measurement instrument, not a failing suite.

Whole solution is green on EF8, EF9 and EF10.

`docs/native-query-status-EF-322.md` is the live status report — the per-slice scoreboard, what is native today, what still falls back, and §9's analysis of what parity would actually require. **Read that before the diff.**

## Reviewing this

Three conventions this branch follows that are worth knowing before reading a test:

1. **MQL shape cannot prove a query went native.** For filter/sort/paging the native and driver pipelines are structurally identical. The only reliable signal is `MongoQueryMode.NativeOnly`: succeeding proves native, throwing `NativeTranslationNotSupportedException` proves a decline.
2. **The oracle is `Native == DriverLinq`.** Where a shape has no driver-LINQ equivalent (cross-collection `SelectMany`, `Intersect`/`Except`, most owned-collection work) tests assert against an in-memory LINQ oracle over the same `Expression` object instead, and say so.
3. **Guards are mutation-verified.** Where a comment claims a guard discriminates, the red count from breaking it is usually recorded beside it.

## Open items a reviewer should know about

**Everything below is ticketed.** Epic EF-322 now has **67** children; 25 of those tickets were filed specifically to park this branch's remaining work.

### Three traps in the code that would mislead a careful reader

- **`TODO(CSHARP-6017)` must not be actioned on the ticket being closed.** CSHARP-6017 is Closed/Done with fixVersion 3.10.0 and this branch pins 3.10.0 — but the tripwire test written to fail when the driver is fixed still **passes**, so the driver still folds a paged join inner into the correlated `$lookup` sub-pipeline and the guard is still load-bearing. Deleting it reintroduces silently wrong join results (measured 0 rows where 453 is correct). The comments have been rewritten to key on the tripwire going red, not on the ticket. **EF-406.**
- **481 `// Fails:` tags reference `EF-X001`–`EF-X020`, which are not JIRA issues.** They are deliberate placeholders documented in `docs/failing-spec-tests.md`; 603 further tags carry real keys. **EF-430** owns replacing them, and carries the per-key counts plus two hazards in the doc itself (`EF-X016` is used for two different subjects; `EF-X004`'s row and heading disagree).
- **Several guards are documented as vacuous or unreachable today**, with mutation evidence in their own comments saying forcing them open turns no test red. They are hardening, not tested protection, and the comments say so — the diff should not be read as adding that much guarded surface.

### Parked work

| Area | Tickets |
|---|---|
| Deferred capability (joins, GroupBy, composite-PK, slice 3d, and friends) | EF-392 … EF-397, EF-247 |
| Committed-but-unstarted plan streams | EF-411 (stream 2, 282 cases), EF-412 (slice 3b, 52 — must *fix* EF-356), EF-413, EF-414, EF-415 |
| Architecture debt the merge plan says needs a decision on record | EF-416 (bulk path / bridge coupling), EF-417 (parity-oracle replacement), EF-418 (`ProjectionAliasTier.Synthetic`) |
| Correctness gaps found and parked, not fixed | EF-407 (deep `ThenInclude` silently-null navigation), EF-408, EF-409, EF-410 |
| Follow-ons and boundaries | EF-419 … EF-425, EF-427, EF-428 |
| Open questions, not known bugs | EF-426 (TPH discriminator narrowing — reconcile with EF-374), EF-429 (sort-key residual under an element `ValueConverter`) |

### Behaviour changes with `BREAKING-CHANGES.md` entries

1. Projecting a required property whose stored element is absent or `null` now throws.
2. A missing or `null` embedded array now materializes as an empty collection, not `null`.
3. Query results can differ for a numeric cast in a `Where` clause.

Each was verified **by execution against the published `v10.0.2` / `v9.1.2` / `v8.4.2` packages**, not inferred from the branch.

## Verifying locally

```bash
# Build one EF version (EF8 / EF9 / EF10)
dotnet build MongoDB.EFCoreProvider.sln -c "Debug EF10"

# Whole solution
dotnet test MongoDB.EFCoreProvider.sln -c "Debug EF10" --no-build

# The coverage axis: which spec tests still need driver-LINQ
MONGODB_EF_NATIVE_ONLY=1 dotnet test \
  tests/MongoDB.EntityFrameworkCore.SpecificationTests/MongoDB.EntityFrameworkCore.SpecificationTests.csproj \
  -c "Debug EF10" --no-build
```

Leave `MONGODB_URI` and `ATLAS_URI` unset — TestContainers then boots a per-process `mongodb/mongodb-atlas-local`, which runs the Atlas-gated tests for real and keeps parallel runs isolated.


src/MongoDB.EntityFrameworkCore/Query/Expressions/ArrayAliasProjectionExpression.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    public INavigation Navigation { get; }
+
+    public Expression AccessExpression { get; }
+
+    public EntityProjectionExpression InnerProjection { get; }
+
+    /// <inheritdoc />
+    /// <remarks>Always <see langword="null"/>: this array is addressed by projection alias.</remarks>
+    public string? ArrayFieldName => null;
```
ajcvickers, line 61: Consider subclasses instead.

src/MongoDB.EntityFrameworkCore/Query/Expressions/ArrayAliasProjectionExpression.cs
```diff
@@ -0,0 +1,91 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
ajcvickers, line 1: Update copyrights

src/MongoDB.EntityFrameworkCore/Query/Expressions/IArrayProjectionExpression.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>The per-element entity projection.</summary>
+    EntityProjectionExpression InnerProjection { get; }
+
+    /// <summary>
+    /// The BSON element name the array sits at, or <see langword="null"/> when the array is addressed by a
+    /// projection alias instead. A <see langword="null"/> here means the caller must already have resolved an
+    /// alias from the owning <see cref="ProjectionExpression"/>; see the <c>??=</c> in
+    /// <see cref="Visitors.MongoProjectionBindingRemovingExpressionVisitor"/>'s <c>VisitBinary</c>.
+    /// </summary>
+    string? ArrayFieldName { get; }
```
ajcvickers, line 42: Consider sub-interfaces.

src/MongoDB.EntityFrameworkCore/Query/Expressions/LookupExpression.cs
```diff
@@ -67,7 +68,10 @@ public LookupExpression(INavigation navigation, bool forceUnwind = false)
     /// <param name="navigation">The navigation the lookup supports.</param>
     /// <returns>The <c>_lookup_&lt;NavigationName&gt;</c> field name.</returns>
     public static string GetLookupAlias(IReadOnlyNavigationBase navigation)
-        => $"_lookup_{navigation.Name}";
+        => $"{LookupAliasPrefix}{navigation.Name}";
+
+    /// <summary>The prefix of the synthetic <c>$lookup</c> alias field (see <see cref="GetLookupAlias"/>).</summary>
+    public const string LookupAliasPrefix = "_lookup_";
```
ajcvickers, line 74: Is this safe?

src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoCardinality.cs
```diff
@@ -0,0 +1,100 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+
+namespace MongoDB.EntityFrameworkCore.Query.Expressions;
+
+/// <summary>The cardinality reducer applied to an entity result.</summary>
+internal enum MongoReducerKind { First, FirstOrDefault, Single, SingleOrDefault }
```
ajcvickers, line 21: Move enums to their own types

src/MongoDB.EntityFrameworkCore/Query/Expressions/MongoConvertExpression.cs
```diff
@@ -0,0 +1,98 @@
...
+/// because a converted-to-<c>null</c> operand then participates in a BSON-total-order comparison
+/// (<c>Null</c> below every number) and quietly moves the row into or out of the result depending on the
+/// operator — reintroducing exactly the silent, operator-dependent behaviour the relational/nullable guard in
+/// <c>MongoExpressionTranslator.CanFallThroughToExpr</c> exists to prevent. A loud abort is the only one of
+/// the three that cannot be mistaken for an answer, and out-of-range data reaching a narrowing cast is a
+/// genuine defect in the query or the data. Recorded in <c>BREAKING-CHANGES.md</c> as an observable delta from
+/// the released packages, with <c>UseQueryMode(MongoQueryMode.DriverLinq)</c> as the mitigation.
+/// </para>
+/// </remarks>
+internal sealed class MongoConvertExpression(MongoExpression operand, Type clrType) : MongoExpression
```
ajcvickers, line 76: This needs some investigation into whether the driver was actually doing the correct thing before.

src/MongoDB.EntityFrameworkCore/Query/NativeTranslation/MongoAggregationExpressionRenderer.cs
```diff
@@ -0,0 +1,186 @@
...
+            MongoFilteredSizeExpression filtered => CanRender(filtered.ElementPredicate),
+            MongoConvertExpression convert
+                => MongoConvertExpression.ToOperatorFor(convert.Type) is not null && CanRender(convert.Operand),
+            _ => false
+        };
+
+    // Exactly the operators RenderBinary's own switch maps below — every MongoBinaryOperator member, as it
+    // happens (RenderBinary has no unmapped member today), but this must be re-checked against RenderBinary's
+    // switch whenever either changes, not assumed to track the enum automatically.
+    private static bool IsRenderableOperator(MongoBinaryOperator op)
```
ajcvickers, line 110: Consider consolidating this into the list that returns the operators so there is only one source of truth.

src/MongoDB.EntityFrameworkCore/Query/Expressions/LookupExpression.cs
ajcvickers: EF-376

PR #328, opened 2026-08-04
EF-370: Correct required-navigation $unwind semantics and stop dropping composed operators

ajcvickers:
Fixes two related correctness defects in cross-collection `Include` translation. They are fixed together because fixing the first alone converts a loud failure into silent wrong data.

### 1. Required reference navigations emitted a left-outer `$unwind`

A required reference navigation emitted `preserveNullAndEmptyArrays: true` where the join's own semantics call for an inner unwind. A dependent row whose foreign key dangles therefore survived the join with a `null` principal instead of being excluded.

The unwind now follows the **LINQ operator** wherever that is observable, falling back to `ForeignKey.IsRequired` only where it is not. `IsRequired` alone was measured insufficient to discriminate the two cases.

### 2. Composed operators silently discarded on a multi-join `Include` (EF-369)

The join `$lookup`s were emitted at the root, above the base source, so any `Where`/`OrderBy`/`Skip`/`Take` composed below them was dropped from the pipeline entirely. The lookups are now emitted above the base source and the composed operators reattached. Ungated by EF version.

### Impact

Both symptoms are **silent wrong data** — wrong rows returned, no exception. Neither has shipped in a released package (the baseline for both is the latest published `v8`/`v9`/`v10` tags, not `main`), so there is no `BREAKING-CHANGES.md` entry.

### Tests

- New `RequiredNavigationUnwindTests` and `Ef369MultiJoinComposedTests`, pinning both `$unwind` shapes. Ungated, so they cover all three EF majors.
- Re-baselines the four Northwind `Include` suites plus `Navigations` and `Miscellaneous` for the inner `$unwind`.
- The previously skipped EF-371 spec failure is baselined as a documented failure rather than skipped (net −1 skip).
- `docs/failing-spec-tests.md` restates only this branch's two rows; the file's totals column does not reconcile under any rule and its basis predates this work, so it was deliberately not patched row-by-row.

Verified: **zero failures on EF8, EF9 and EF10.**

### Follow-ups filed, not included

- **EF-372** — a 3-hop `ThenInclude` emits an unprefixed `localField` at hop 3, so the inner `$unwind` drops every row.
- **EF-373** — `Skip`/`Take` interleaved *between* two joins is still hoisted above both lookups; this PR fixes only the below-the-joins case.


damieng (changes requested):
Need to drop the superpowers specs etc.

damieng (approved):

PR #335, opened 2026-08-12
EF-352: Fix shadow property via EF.Property returning null in join projections

ajcvickers:
TryBindNavigationMemberAccess only recognised plain MemberExpression access on a joined entity's shaper. Shadow properties have no CLR member and can only be read via EF.Property, so they fell through to the root-document field lookup and silently materialised as null. Extend the binder to also match EF.Property calls on the joined shaper, resolving the property by name instead of by MemberInfo.


damieng:
Fixed in #327

PR #337, opened 2026-08-12
EF-357: Fix ArgumentException on bare embedded-collection Count projections

ajcvickers:
db.Set<T>().Select(e => e.OwnedCollection.Count) threw ArgumentException in every case (bare and arithmetic-wrapped alike), because EF Core lowers a collection-navigation Count to Queryable.Count(Queryable.AsQueryable(source)) even for an in-document embedded List<T>, and
MongoProjectionBindingExpressionVisitor had no case for that shape - it fell into the generic fallback, which tried to rebuild the Queryable.Count call with the already-materialized List<T>, which isn't an IQueryable<T>.

Add TryBindEmbeddedCollectionNavigationCount to recognize the shape and rewrite it as a null-safe Enumerable.Count/LongCount over the navigation's CollectionShaperExpression (null-safe because a missing/explicitly-null stored array materializes to null rather than an empty collection).


ajcvickers:
Superseded by https://github.com/mongodb/mongo-efcore-provider/pull/339

PR #338, opened 2026-08-12
EF-358: a missing or null embedded array materializes as an empty collection

ajcvickers:
The driver-LINQ collection shaper left a missing or explicitly-BSON-null stored owned-collection array as a null CLR value: BsonDocumentInjectingExpressionVisitor collapsed the whole CollectionShaperExpression to a null constant, and MongoProjectionBindingRemovingExpressionVisitor never coalesced the raw BsonArray read before enumerating it. Because whole-entity materialization, Include, and projections all compile through this same shaper, every one of them was affected - what looked like a working whole-entity path was really each POCO's own field initializer papering over the null.

Removes the null-collapsing condition and coalesces the BsonArray read to a fresh empty array at its point of use (kept there rather than folded into the assignment, since VisitBinary hard-casts that assignment's right-hand side to UnaryExpression). A collection navigation now uniformly materializes empty, matching EF Core's contract, and a projected .Count/.Any() on such a collection no longer throws.

Native/streaming shader materialization and the Count-projection translation fix are out of scope here - the native query pipeline isn't on this branch, and Count-projection translation is a separate, unimplemented effort.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -4,8 +4,6 @@ Please note that this provider **does not follow traditional semantic versioning
 
 In order to evolve the provider as we introduce new features, we will be using the minor version number for breaking and significant changes to our EF Core provider. Please bear this in mind when upgrading to newer versions of the MongoDB EF Core Provider and ensure you read the release notes and this document for the latest in breaking change information.
 
-## Breaking changes in 8.5.0 / 9.2.0 / 10.1.0
-
 ### Entity types with their own `DbSet` are no longer embedded when reached by a navigation
 
 #### Old behavior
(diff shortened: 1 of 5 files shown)


BREAKING-CHANGES.md
```diff
@@ -6,6 +6,61 @@ In order to evolve the provider as we introduce new features, we will be using t
 
 ## Breaking changes in 8.5.0 / 9.2.0 / 10.1.0
 
+### A missing or `null` embedded array now materializes as an empty collection, not `null`
```
ajcvickers: Claude wrote this. Not sure if it is needed or not; this could be considered a simple bug fix.

damieng: I'd probably not include it but can go either way.

damieng (changes requested):
I think we need to massively prune that breaking change doc. It's a life story for a simple behavioral change.  We also need to reduce the amount of code comments Claude is emitting in our code base.

ajcvickers:
I'll remove the breaking change note. I'm in two minds about the comments, since they are now relevant to agents more than humans, so I'm slightly leaning towards allowing them to get more verbose than normal. That being said, I have Claude condense the comments in a big PR and that made them better, so I can do the same here.

PR #340, opened 2026-08-13
EF-360: fix ArgumentException projecting a collection leaf whose element has a nested embedded navigation

ajcvickers:
An embedded (OwnsMany) collection whose element type has its own navigation (e.g. a nested OwnsMany/OwnsOne) crashed when projected as an anonymous-type/MemberInit leaf, e.g. Select(b => new { b.Title, b.Posts }) where Post has its own Comments collection.

EF's nav-expansion wraps the collection access in a Queryable.Select carrying the auto-included nested navigation; MongoProjectionBindingExpressionVisitor's Select arm rebuilds that against EnumerableMethods.Select, producing an IEnumerable<T>-typed result instead of the navigation's declared List<T>. MatchTypes deliberately skips collection-typed targets, so the mismatch reached Expression.New's member-type validation as an ArgumentException instead of failing (or succeeding) cleanly.

Fix: in the embedded-navigation branch of the MaterializeCollectionNavigationExpression case, wrap the visited subquery in a Convert to the navigation's declared type when it differs. This is a no-op at runtime: MongoProjectionBindingRemovingExpressionVisitor already special-cases this exact Select-over-IncludeExpression shape later in the pipeline and discards it for the correctly-typed CollectionShaperExpression.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
@@ -203,17 +203,11 @@ internal sealed partial class MongoProjectionBindingExpressionVisitor : Expressi
                 {
                     var visited = base.Visit(materializeCollectionNavigationExpression.Subquery);
 
-                    // When an element of the collection has an embedded navigation of its own (e.g. a nested
-                    // OwnsMany/OwnsOne), EF's nav-expansion wraps the collection access in a Queryable.Select
-                    // carrying the auto-included IncludeExpression. The Select arm above rebuilds that against
-                    // EnumerableMethods.Select, which is IEnumerable<T>-typed, not the navigation's declared
-                    // List<T>-typed collection. MatchTypes deliberately leaves collection-typed targets alone
-                    // (see its TryGetItemType() guard), so left uncorrected this fails Expression.New's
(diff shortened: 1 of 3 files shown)


src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoProjectionBindingExpressionVisitor.cs
```diff
@@ -161,10 +161,28 @@ protected override Expression VisitExtension(Expression extensionExpression)
...
+                    // OwnsMany/OwnsOne), EF's nav-expansion wraps the collection access in a Queryable.Select
+                    // carrying the auto-included IncludeExpression. The Select arm above rebuilds that against
+                    // EnumerableMethods.Select, which is IEnumerable<T>-typed, not the navigation's declared
+                    // List<T>-typed collection. MatchTypes deliberately leaves collection-typed targets alone
+                    // (see its TryGetItemType() guard), so left uncorrected this fails Expression.New's
+                    // member-type validation wherever this leaf feeds an anonymous type / MemberInit member.
+                    // MongoProjectionBindingRemovingExpressionVisitor.VisitMethodCall already discards this
+                    // exact Select-over-IncludeExpression shape and hands back the properly List<T>-typed
+                    // CollectionShaperExpression, so wrapping it here in a Convert is a no-op by the time it's
+                    // actually consumed - it only exists to satisfy the static-type check at this stage.
```
damieng: I think we might want to rein in Claude's verbosity on code comments.

damieng (approved):

PR #350, opened 2026-08-14
EF-376: Fix sibling ThenIncludes through the same target type collapsing into one $lookup

ajcvickers:
Two sibling reference navigations sharing a target type (e.g. Root.PrimaryMid.Leaf and Root.SecondaryMid.Leaf both reaching "Mid.Leaf") resolved their transitive $lookup's "through" join by target entity type alone, which can't distinguish the two paths. Both ended up aliased "_lookup_Leaf" and the second was silently dropped by the alias-dedup in AddLookup, so one branch never materialized.

Resolve the "through" join positionally instead, by walking the outer key selector's .Outer/.Inner chain, and only fall back to a path-qualified alias when a different intermediate has already claimed the plain name - keeping existing non-ambiguous transitive joins' aliases (and $lookup emission order) unchanged.


ajcvickers:
@damieng Note that this is now just new testing, since the functional fix was fully contained by the already merged work.

damieng (approved):

PR #353, opened 2026-08-18
EF-379: Fix transitive ThenInclude hop misrouted to root-level decoy navigation

ajcvickers:
RebindInnerShaperToOuterQuery matched a join's navigation by FK-property name or target-entity-type alone against the root entity's navigations. When a transitive ThenInclude hop's FK is actually read from an intermediate joined entity, but the root coincidentally has a same-named FK or a navigation to the same target type, the match fired against the root instead, producing a $lookup on the wrong field and a silently null navigation.

Add IsRootLevelKeyAccess, which inspects the FK selector's receiver and peels EF's TransparentIdentifier Outer/Inner chain back to the join lambda parameter; any Inner hop means the access is transitive, so the root-navigation match is skipped and the existing intermediate-resolution branch runs instead.

Driver-based LINQ path only; the native LINQ write branch already carries its own fix for the analogous issue.


ajcvickers:
@damieng Note that this is now just new testing, since the functional fix was fully contained by the already merged work.

damieng (approved):

PR #361, opened 2026-08-18
EF-228: Fix Average/Sum over float throwing on precision loss

ajcvickers:
MongoDB always computes $avg/$sum in double precision. When the LINQ selector is float-typed, the driver deserializes the result with the strict SingleSerializer, which throws TruncationException rather than narrow a double that isn't exactly representable as a float.

Rewrite the captured Average/Sum call to the widening double overload before handing it to the driver, and narrow the result back to float client-side (a plain numeric conversion, not a BSON round-trip) once execution completes.

### DIFF
--- a/docs/failing-spec-tests.md
+++ b/docs/failing-spec-tests.md
@@ -46,7 +46,7 @@ that currently lack a ticket. Counts are sourced from `tests/MongoDB.EntityFrame
 | [EF-221](https://jira.mongodb.org/browse/EF-221) | `Equals with different types issue EF-221` | `==` / `Equals` with operands of mismatched CLR types (e.g. `int == long`) is not translated correctly. | 4 |
 | [EF-222](https://jira.mongodb.org/browse/EF-222) | `translation of Like issue EF-222` | `EF.Functions.Like(...)` is not translated. | 9 |
 | [EF-227](https://jira.mongodb.org/browse/EF-227) | `Max over empty nullables issue EF-227` | `Min` / `Max` over an empty nullable sequence does not produce the EF-expected `null`. | 4 |
-| [EF-228](https://jira.mongodb.org/browse/EF-228) | `Truncation data loss issue EF-228` | `Sum`/`Average` over `float` columns suffers precision/truncation loss when accumulated server-side. | 2 |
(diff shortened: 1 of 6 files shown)


damieng (commented):
This one feels very specific to Average/Sum.  Are we sure this is only in this scenario and doesn't cover things like div etc?  This might be a broader pattern?

ajcvickers:
@damieng Good question — I dug into this before merging, and it turns out arithmetic (`+`, `-`, `*`, `/`) is *not* exposed to the same failure mode, so the fix doesn't need to generalize there.

Root cause is actually a quirk isolated to the driver's `Average`/`Sum` translators, not something general to "MongoDB computes in double":

- Plain arithmetic inside a `Select` gets its result deserializer from the driver's `StandardSerializers` table (`Linq3Implementation/Serializers/StandardSerializers.cs`), which registers `SingleSerializer` with an `allowTruncationConverter` — i.e. it's already lenient about narrowing double → float.
- `Average`/`Sum`, however, are hardcoded in the driver's own translators (`AverageMethodToExecutableQueryTranslator.cs` and the Sum equivalent) to resolve the output serializer to the *strict* `SingleSerializer.Instance` (no truncation allowed) for a `float` result type. That's the one that throws `TruncationException`.

I verified this empirically: ran ~12 float division/multiplication pairs chosen specifically to lose precision on the double→float narrow (e.g. `0.1f / 3.0f`, `1f / 3f`, `22f / 7f`), and none of them throw — confirmed via MQL logging that the division genuinely happens server-side (`$divide`), so it's not a client-side-eval accident.

I also checked the other candidates you might be thinking of:
- `GroupBy` is rejected by this provider outright (`TranslateGroupBy` returns `null` unconditionally), so a grouped `g.Average(...)`/`g.Sum(...)` never reaches a `$group` accumulator through this provider today.
- `Min`/`Max` just select an existing stored value rather than compute a new one, so there's no newly-computed double to narrow.

So the exposure really is Average/Sum-specific, and the fix as scoped is complete. Happy to add a note to the PR description/docs capturing this if useful.

ajcvickers:
@damieng Claude investigated, but this is appropriately scoped for now. Full comment above from Claude.

damieng:
This feels like it's a bug that should be fixed in the driver then.  At least allowing truncation in the sum/average but ideally in both those places being smarter about using double.

ajcvickers:
This is better deferred to driver or native LINQ implementation.

PR #363, opened 2026-08-18
EF-233: Fix compiled query ending in a non-query operator

ajcvickers:
MongoQueryableMethodTranslatingExpressionVisitor captured the "final" query-chain expression in its generic Visit override, which fires on whatever node is visited first. For a compiled query ending in a non-query operator (e.g. Where(...).Count() == 1), the wrapping BinaryExpression was visited before the actual Count() call, so the captured expression included the trailing comparison and downstream translation broke.

Move the capture into VisitMethodCall, which only ever fires on the true queryable-chain nodes, so the outermost method call is captured regardless of any non-queryable wrapper around it.

### DIFF
--- a/docs/failing-spec-tests.md
+++ b/docs/failing-spec-tests.md
@@ -90,7 +90,7 @@ These entries appear in `// Fails:` comments without an `EF-` or `CSHARP-` refer
 | EF-X008 | No support for nested JSON in AdHoc JSON tests | 2 |
 | EF-X009 | Single uncategorized failure — needs triage | 1 |
 | EF-X010 | Provider-specific Include error message differs from EF baseline | 4 |
-| EF-X011 | Compiled query with non-query operator — wrong exception | 1 |
+| EF-X011 | ~~Compiled query with non-query operator — wrong exception~~ — fixed by EF-233 | 0 |
 | EF-X012 | `OfType` translation unsupported | 2 |
 | EF-X013 | MongoDB has no `$xor` operator (`Where_bitwise_xor`) | 1 |
 | EF-X014 | Server-side projection conflict with cast-to-nullable | 1 |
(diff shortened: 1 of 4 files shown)


ajcvickers:
Follow-up on the `query-reviewer` flag re: `NorthwindCompiledQueryMongoTest.Multiple_queries`.

Before this PR, that test's unsupported cross-DbSet compiled query hit the provider's own `InvalidOperationException("Unsupported cross-DbSet query between...")` guard (the `EntityQueryRootExpression` mismatch check in `MongoEFToLinqTranslatingExpressionVisitor.Visit`). After this PR it surfaces a generic `ArgumentException` from `Expression.Call`'s argument-type validation instead, bypassing that guard — still a failure either way, just a less actionable message.

Tracing it further: the `ArgumentException` happens because `base.VisitMethodCall` tries to reconstruct `OrderBy(Customer, ...)` with an `Order`-typed source already substituted in — i.e. something fed it the wrong-typed source *before* the guard's `EntityQueryRootExpression` node was ever visited. That points to visitor-instance state (e.g. `_foundEntityQueryRootExpression`) leaking across the two independently-`EF.CompileQuery`'d delegates in that test, rather than an actual cross-DbSet reference within a single query tree. That's a different mechanism than what's tracked under EF-216 (cross-collection navigation within one query) — it looks more like `MongoQueryableMethodTranslatingExpressionVisitor` / `MongoEFToLinqTranslatingExpressionVisitor` instances being incorrectly reused across separate compiled-query translations.

Recommend filing this as its own follow-up ticket (distinct from EF-216) rather than folding it in — the fix likely involves auditing translator-visitor lifetime/scoping for compiled queries, which is out of scope for EF-233's `_finalExpression` capture-timing fix.

damieng (approved):

PR #367, opened 2026-08-21
EF-254: Fix Take(0) throwing ArgumentOutOfRangeException

ajcvickers:
Take(0) (constant or parameterized) flowed straight through to the driver's AstLimitStage, which rejects a limit of 0 since a $limit stage of 0 is meaningless server-side. Rewrite Take(0) to an equivalent Where(_ => false) during EF-to-driver-LINQ translation instead, so the provider never asks the driver for $limit: 0.

### DIFF
--- a/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
+++ b/src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
@@ -616,7 +616,42 @@ internal sealed partial class MongoEFToLinqTranslatingExpressionVisitor : System
             return sizeRewrite;
         }
 
+        var zeroTakeRewrite = TryRewriteZeroTake(node);
+        if (zeroTakeRewrite != null)
+        {
+            return zeroTakeRewrite;
+        }
+
         return base.VisitMethodCall(node);
+    }
+
+    /// <summary>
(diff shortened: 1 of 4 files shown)


src/MongoDB.EntityFrameworkCore/Query/Visitors/MongoEFToLinqTranslatingExpressionVisitor.cs
```diff
@@ -616,9 +616,44 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+    /// Rewrites <c>source.Take(0)</c> (constant or parameterized) into <c>source.Where(_ => false)</c>.
+    /// The driver's <c>AstLimitStage</c> rejects a limit of 0 (EF-254) — a MongoDB <c>$limit</c> stage of 0
+    /// is meaningless server-side, so the driver's guard is correct and shouldn't be relaxed. The provider
+    /// already knows the concrete count by translation time (EF query parameters are resolved to constants
+    /// upstream of this visitor), so it can short-circuit to the equivalent empty-result shape itself
+    /// without ever emitting <c>$limit</c>.
+    /// </summary>
+    private Expression? TryRewriteZeroTake(MethodCallExpression node)
+    {
+        if (node.Method.Name != nameof(Queryable.Take)
```
damieng, line 638: We normally have the detect-if-I-should-call-this at the level above the call across LINQ presumably to prevent unnecessary calls.  I'm not sure this is much of a perf benefit these days though given all the work on .NET optimization - might be worth benchmarking at some point so we have a guide on which way to go.

damieng (approved):

